PR #21, opened 2018-12-16
GoogleMapPersitedInstace Component - Google Maps Memory Leak  

xalisys:
* Fixed Google Maps Memory Leak mentioned in [here](https://stackoverflow.com/questions/21142483/google-maps-js-v3-detached-dom-tree-memory-leak)
* Updated Docs and examples.


src/GoogleMap.js
```diff
@@ -61,23 +61,30 @@ const updaterMap = {
     map.setZoom(zoom)
   }
 }
-
+const defatulMapId = "defaultMapId";
```
uriklar: typo :-)

src/GoogleMap.js
```diff
@@ -61,23 +61,30 @@ const updaterMap = {
...
-    onLoad: () => {}
+    onLoad: () => { }
   }
-
-  state = {
-    map: null
+  get mapId() {
+    return this.props.id || defatulMapId;
+  }
+  constructor(props) {
```
uriklar: We're using class properties here. So no need for a constructor. state can stay outside as it was and getInstance can be written as an arrow function (instead of binding)

src/GoogleMap.js
```diff
@@ -61,23 +61,30 @@ const updaterMap = {
...
   static propTypes = GoogleMapPropTypes
   static defaultProps = {
-    onLoad: () => {}
+    onLoad: () => { }
   }
-
-  state = {
-    map: null
+  get mapId() {
+    return this.props.id || defatulMapId;
```
uriklar: Please use defaultProps instead of this constant

src/GoogleMap.js
```diff
@@ -104,7 +111,7 @@ export class GoogleMap extends PureComponent {
     })
   }
 
-  componentWillUnmount = () => {
```
uriklar: Any reason for changing this?

src/GoogleMap.js
```diff
@@ -114,6 +121,7 @@ export class GoogleMap extends PureComponent {
 
   render = () => (
     <div
+      id={this.mapId}
```
uriklar: after using defaultProps this could just be ```this.props.id```

src/GoogleMapPersistedInstance.js
```diff
@@ -0,0 +1,58 @@
...
+
+const clearChildren = (node) => {
+  if (node) {
+    while (node.firstChild) {
+      node.removeChild(node.firstChild);
+    }
+  }
+}
+
+export class GoogleMapPersistedInstance extends GoogleMap {
```
uriklar: Could this functionality become a part of the ```GoogleMap``` component? 
Is there a case where i'd want to use GoogleMap without the persisted instance functionality?

xalisys:
@uriklar  @JustFly1984  I pushed new fixes please have look, thanks for feedback.

JustFly1984:
I've created an issue https://issuetracker.google.com/issues/121054556 in google bug tracker

PR #22, opened 2018-12-18
Using cross-env instead win-node-env.

gregoryfm:
When I run the yarn command, I get this message:
Unsupported platform for win-node-env@0.4.0: wanted {"os":"win32","arch":"any"} (current: {"os":"darwin","arch":"x64"})

Replacing win-node-env for cross-env because win-node-env don't work in linux/macOS environment.
cross-env was already installed but was not being used.


package.json
```diff
@@ -66,8 +66,8 @@
     "watch": "npm-watch build:dev",
     "dtslint": "dtslint types",
     "test": "npx jest",
-    "build": "concurrently \"NODE_ENV=production npm run build:webpack\" \"NODE_ENV=production npm run build:babel\"",
-    "build:dev": "concurrently \"NODE_ENV=development npm run build:webpack\" \"NODE_ENV=development npm run build:babel\"",
+    "build": "concurrently \"cross-env NODE_ENV=production yarn build:webpack\" \"cross-env NODE_ENV=production yarn build:babel\"",
```
uriklar: @gregoryfm any specific reason to replace npm here with yarn?

uriklar:
@xalisys win-node-dev indeed breaks yarn install for me too. Can you please confirm that you can run the build steps using @gregoryfm's fix (cross-env)?

package.json
gregoryfm: Sorry about that. I always use yarn and I replaced in automatic mode, I didn't think about it. :)
I'll return to npm.

PR #23, opened 2018-12-18
Add polygon example to docz

gregoryfm:


JustFly1984:
oh it is merged, never-mind)

PR #26, opened 2019-01-12
Prevent mounting LoadScript while we are unmounting previous one

anton-nagornyi:
Always delete windows.google on unmount

### DIFF
--- a/src/LoadScript.js
+++ b/src/LoadScript.js
@@ -1,4 +1,4 @@
-import React,{ Component } from 'react'
+import React, { Component } from 'react'
 import { injectScript } from './utils/injectscript'
 import { LoadScriptPropTypes } from './proptypes'
 import { preventGoogleFonts } from './utils/prevent-google-fonts'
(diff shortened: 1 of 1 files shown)


uriklar:
@anton-nagornyi @JustFly1984 Hi guys!
Question: What if, instead of all these checks for if we are cleaning up, we simply always perform a cleanup on componentDidMount. Would that work? 
The only problem I can think of is multiple LoadScript tags, but do we even want to allow multiple load script tags? Is there a use case where we want to load the google maps script more than once?

anton-nagornyi:
1. The whole point of cleaning up here is to do "delete window,google". You can't do that onComponentWillUnmount because some of a children components may rely on window.google existence while unmounting. There is a case when map's language is changed and then we are also doing that cleanup. Previously we removed window.google when LoadScript state was changed, and for the case of switching language it was ok. But that cleaning up never fired when component was unmounting, so I guess it was a bug because we had window.google object stayed after LoadScript was unmounted.

2. I don't think we need to support case with several LoadScript components active at the same time, but there is a case while we are unmounting first one and mounting another one. And we need to make sure that cleanup was really finished before mounting new LoadScript

JustFly1984:
Well, we just can’t support more than one LoadScript in the app. In previous version it was not even possible to have multiple maps, so that was my goal

anton-nagornyi:
And now we still can't

PR #27, opened 2019-01-13
Added lerna

anton-nagornyi:


JustFly1984:
Well, it is failing, cos we don't have PR branch deployment, so don't bother about it.

JustFly1984:
Sorry I don't have energy to go through commit today, will try to find the time tomorrow. @uriklar maybe you can find a time to review?

anton-nagornyi:
@JustFly1984 Checkout CONTRIBUTING.md

@uriklar 
We started to use lerna because we wanted to get rid of the peer dependency hell. Storybook is conflicting with docz on @emotion/core dependency. Now those projects are separated and conflicts are resolved. Also I think it is a good idea to keep things separate )

packages/react-google-maps-api/src/components/streetview/StreetViewPanorama.js
```diff
@@ -73,13 +73,12 @@ export class StreetViewPanorama extends PureComponent {
 
   componentDidMount = () => {
     const streetViewPanorama = this.context.getStreetView()
-
     this.setState(
       () => ({
         streetViewPanorama
       }),
       () => {
-        this.state.streetViewPanorama.setVisible(true)
```
uriklar: Is this change intentional?

packages/docz/src/Marker.mdx
```diff
@@ -2,11 +2,7 @@
 name: Marker
 ---
 
-import {
-  GoogleMap,
-  LoadScript,
-  Marker
-} from "..";
+import { GoogleMap, LoadScript, Marker } from "react-google-maps-api/lib";
```
uriklar: Why are we using the /lib here?

packages/react-google-maps-api/src/components/streetview/StreetViewPanorama.js
anton-nagornyi: Not really. StreetViewPanorama is not finished yet and is not available for use. I made some changes to it before lerna and they went to PR accidentally

PR #35, opened 2019-01-25
Fixed bug in render SearchBox component.

awwit:


uriklar:
Hi @awwit, I think it would be great if we could have a usage example. I personally am not sure how to use this prop..
Could you maybe add a docz page for this component?

awwit:
@uriklar 
https://github.com/JustFly1984/react-google-maps-api/blob/master/src/components/places/SearchBox.js

In the component is not assigned anywhere: `this.state.containerElement`.
`containerElement` is in the `props` of the component, not in its `state`.

awwit:
@uriklar I can not give an example, but there definitely should be: `props`.
Only now, I saw that this component is not completed and not used.

---

Found examples of using api:
1. react-google-maps: https://tomchentw.github.io/react-google-maps/#searchbox
2. Google Maps Docs: https://developers.google.com/maps/documentation/javascript/examples/places-searchbox

PR #41, opened 2019-02-04
Typescript re-write

uriklar:
Rewrite all finished components using typescript
Replace webpack/babel build with typescript build


packages/react-google-maps-api/src/components/directions/DirectionsRenderer.tsx
```diff
@@ -8,49 +8,52 @@ import {
...
 
   componentDidMount = () => {
     const directionsRenderer = new google.maps.DirectionsRenderer(
-      Object.assign(
-        {
-          map: this.context
-        },
-        this.props.options
-      )
+      this.props.options
```
JustFly1984: Have you tested this change? my project heavily depends on this component. I do not really want to pass map to options object from outside. I do remember the reason I've written this line of code :)

packages/react-google-maps-api/src/components/drawing/Data.tsx
```diff
@@ -0,0 +1,160 @@
+/* global google */
```
JustFly1984: How many `<Data />` components can we render at the same time?

packages/react-google-maps-api/src/components/drawing/Polygon.tsx
```diff
@@ -25,54 +24,67 @@ const eventMap = {
...
 
   componentDidMount = () => {
     const polygon = new google.maps.Polygon(
-      Object.assign(
-        {
-          map: this.context
-        },
-        this.props.options
-      )
+      this.props.options
```
JustFly1984: Here is the same question. I do not want to pass map to options through jsx

packages/react-google-maps-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@react-google-maps/api",
-  "version": "1.0.8",
+  "version": "1.1.0",
```
JustFly1984: You do not need to update package version. Now I do it with `lerna publish` in root, and it bumps version automagically

packages/react-google-maps-api/src/components/drawing/InfoWindow.tsx
```diff
@@ -20,46 +20,48 @@ const eventMap = {
...
   }
 
   componentDidMount = () => {
     const infoWindow = new google.maps.InfoWindow(
-      Object.assign(
-        {
-          map: this.context
-        },
-        this.props.options
-      )
```
JustFly1984: here is the same question

packages/react-google-maps-api/src/components/drawing/InfoWindow.tsx
```diff
@@ -96,20 +98,18 @@ export class InfoWindow extends PureComponent {
 
   componentWillUnmount = () => {
     unregisterEvents(this.registeredEvents)
-
-    this.state.infoWindow && this.state.infoWindow.setMap(null)
```
JustFly1984: Should not we cleanup?

packages/react-google-maps-api/src/components/fusion/FusionTablesLayer.tsx
```diff
@@ -8,40 +8,40 @@ import {
...
   }
 
   componentDidMount = () => {
     const fusionTablesLayer = new google.maps.FusionTablesLayer(
-      Object.assign(
-        {
-          map: this.context
-        },
-        this.props.options
-      )
```
JustFly1984: Here is the same question :)

packages/react-google-maps-api/src/components/kml/KmlLayer.tsx
```diff
@@ -4,45 +4,49 @@ import { PureComponent } from 'react'
...
   }
 
   componentDidMount = () => {
     const kmlLayer = new google.maps.KmlLayer(
-      Object.assign(
-        {
-          map: this.context
-        },
-        this.props.options
-      )
```
JustFly1984: here is the same question : )))

packages/react-google-maps-api/src/components/maps/BicyclingLayer.tsx
```diff
@@ -3,26 +3,15 @@ import { PureComponent } from 'react'
...
 
   componentDidMount = () => {
-    const bicyclingLayer = new google.maps.BicyclingLayer(
-      Object.assign(
-        {
-          map: this.context
-        },
-        this.props.options
-      )
-    )
```
JustFly1984: And here ))))

packages/react-google-maps-api/src/components/maps/TrafficLayer.tsx
```diff
@@ -7,33 +7,34 @@ import {
...
+  registeredEvents: google.maps.MapsEventListener[] = []
+
   componentDidMount = () => {
     const trafficLayer = new google.maps.TrafficLayer(
-      Object.assign(
-        {
-          map: this.context
-        },
-        this.props.options
-      )
```
JustFly1984: And here ))))))

packages/react-google-maps-api/src/components/overlays/GroundOverlay.tsx
```diff
@@ -16,40 +16,47 @@ const eventMap = {
...
     )
   }
 
   componentDidMount = () => {
     const groundOverlay = new google.maps.GroundOverlay(
       this.props.url,
       this.props.bounds,
-      Object.assign({
-        map: this.context
-      },
```
JustFly1984: )))))))

packages/react-google-maps-api/src/components/places/StandaloneSearchBox.tsx
```diff
@@ -1,38 +1,45 @@
 /* global google */
-import React, {
-  PureComponent,
-  Children
-} from 'react'
+import * as React from 'react'
```
JustFly1984: IMHO for me explicit imports tells a lot about file functionality. I do not like `import *` it is better to use

```
import {
  default as React,
  PureComponent,
  Children
} from 'react'

File: packages/react-google-maps-api/src/components/places/StandaloneSearchBox.tsx [hunk 13]
```diff
@@ -42,12 +49,7 @@ class StandaloneSearchBox extends PureComponent {
   componentDidMount = () => {
     const searchBox = new google.maps.places.SearchBox(
       this.containerElement.current.querySelector('input'),
-      Object.assign(
-        {
-          map: this.context
-        },
-        this.props.options
-      )
```
JustFly1984: )))))

packages/react-google-maps-api/src/GoogleMap.tsx
```diff
@@ -141,18 +180,20 @@ export class GoogleMap extends PureComponent {
...
 
-  panToBounds = (...args) => this.state.map.panToBounds(...args)
+  panToBounds = (latLngBounds: Bounds, padding?: number | google.maps.Padding) => this.state.map.panToBounds(latLngBounds, padding)
 
   getBounds = () => this.state.map.getBounds()
 
   getCenter = () => this.state.map.getCenter()
 
+  // Function exists but it missing in typings
+  //@ts-ignore
```
FredyC: No need to split into two lines :)
```suggestion
  //@ts-ignore Function exists but it missing in typings
```

packages/react-google-maps-api/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "compilerOptions": {
+    "outDir": "./lib/",
+    "sourceMap": true,
+    "noImplicitAny": false,
```
FredyC: Too many `any` in overall? :) Wondering if that could be improved, haven't checked all cases yet.

packages/react-google-maps-api/src/components/drawing/Polygon.tsx
uriklar: You don't pass the map prop through jsx. It is passed via context

packages/react-google-maps-api/package.json
uriklar: I see. And do you tell it to bump a major/minor or patch? Because I think this should be a minor version

packages/react-google-maps-api/src/components/drawing/InfoWindow.tsx
uriklar: InfoWindow doesn't have a ```setMap``` method. The dom node is remove when ```infoWindow.close()``` is triggered (that happens by clicking the X and currently we are not involved in that process)

packages/react-google-maps-api/src/components/places/StandaloneSearchBox.tsx
uriklar: Fixed (and in 3 other components with same issue)

packages/react-google-maps-api/src/components/drawing/Data.tsx
uriklar: I'm actually just seeing now that we don't export Data in index.ts. We need to add an example/docz page for it. Why do you ask? I don't see any constraint on the number of Data components that can be rendered

packages/react-google-maps-api/src/GoogleMap.tsx
```diff
@@ -28,50 +28,91 @@ const eventMap = {
...
+  onHeadingChanged?: () => void;
+  onIdle?: () => void;
+  onProjectionChanged?: () => void;
+  onResize?: () => void;
+  onTiltChanged?: () => void;
+  onZoomChanged?: () => void;
+  onLoad?: (map: google.maps.Map) => void;
+}
+
+export class GoogleMap extends React.PureComponent<GoogleMapProps, GoogleMapState> {
```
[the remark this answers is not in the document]
uriklar: I think they will still get useful prop info in their ide even if they don't use typescript (at least in vscode for sure). Regarding PropTypes. I'm not sure how it works right now, but don't they get stripped at build time anyway?

packages/react-google-maps-api/tsconfig.json
uriklar: My main use case for any is in:
https://github.com/JustFly1984/react-google-maps-api/blob/typescript/packages/react-google-maps-api/src/utils/helper.ts

The functions there receive a component, it's props, it's updaterMap object and it's eventMap object. These are different for each component.

FredyC: Well, in that case, it might be better to ignore per case instead of turning that guard off everywhere and possibly missing some cases.

packages/react-google-maps-api/src/GoogleMap.tsx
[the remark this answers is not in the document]
FredyC: Yea, propTypes are DEV only and removed in prod builds. It's a bit different as it checks in runtime instead of compile time. Original react-google-maps does not care about propTypes either, so let's forget about it :)

packages/react-google-maps-api/src/components/drawing/Rectangle.tsx
```diff
@@ -49,28 +49,35 @@ interface RectangleState {
...
+  onRightClick?: (e: MouseEvent) => void;
+  onClick?: (e: MouseEvent) => void;
+  onDrag?: (e: MouseEvent) => void;
+  onBoundsChanged?: () => void;
 }
 
 export class Rectangle extends PureComponent<RectangleProps, RectangleState> {
   static contextType = MapContext
 
+  static defaultProps: Partial<RectangleProps> = {
```
FredyC: Uhh, I don't get it. Why would you turn every prop optional in the declaration above and then used the `Partial` here? It's the same thing :)

uriklar: I turned them optional because they were required by accident. What you're saying is that if they are all optional there's no need for the Partial right? And I think you are right :-)

packages/react-google-maps-api/src/components/maps/BicyclingLayer.tsx
uriklar: Bicycling layer constructor doesn't accept options property

packages/react-google-maps-api/src/components/places/StandaloneSearchBox.tsx
uriklar: Search box doesn't accept map in constructor or in options

packages/react-google-maps-api/package.json
```diff
@@ -158,18 +114,15 @@
...
     "react-dom": "16.7.0",
+    "react-styleguidist": "^8.0.6",
     "rimraf": "2.6.3",
     "standard-version": "4.4.0",
-    "terser-webpack-plugin": "1.2.1",
-    "typescript": "3.3.1",
-    "webpack": "4.29.0",
-    "webpack-bundle-analyzer": "3.0.3",
-    "webpack-cli": "3.2.1"
+    "typescript": "^3.2.4",
```
asaarnak: 3.3.3 is current latest

PR #58, opened 2019-03-07
Move @types/googlemaps to be a direct dependency

### LINKED ISSUE #57: Should @types/googlemaps be in dependencies?
## Your Environment

(other details unlikely to be relevant)

@react-google-maps/api version = 1.1.0

## How is it behave?

> How **does** it behave :)

In the exported Typescript definitions we have an old-fashioned reference comment, such as:

```
/// <reference types="googlemaps" />
```
(from `@react-google-maps/api/lib/components/drawing/Marker.d.ts`)

and further down the same file we have references to the Google Maps types:

```
interface MarkerProps {
    options?: google.maps.MapOptions;
    animation?: google.maps.Animation;
...
```

The reference is to `@types/googlemaps` but as this is in devDependencies it is necessary to track this down and manually install, so that you can refer to the types within the IDE.

If `@types/googlemaps` is in dependencies it will be installed for consumers of the package as well. There could be arguments against this, such as package bloat, but would give a better experience for users IMO.

robations:
Fixes #57 so that consumers have access to googlemaps types in IDE.

### DIFF
--- a/packages/react-google-maps-api/package.json
+++ b/packages/react-google-maps-api/package.json
@@ -66,6 +66,7 @@
     "docs:build": "npx styleguidist build"
   },
   "dependencies": {
+    "@types/googlemaps": "3.30.16",
     "google-maps-infobox": "2.0.0",
     "invariant": "2.2.4",
     "marker-clusterer-plus": "2.1.4",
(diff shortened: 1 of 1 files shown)


FredyC:
You should probably use Yarn to update `yarn.lock` as well. Unless maintainers want to do that by themselves which is generally safer :)

PR #69, opened 2019-03-26
alpha 5

JustFly1984:
added new eslint plugin, added optional onUnmount prop for each component. improved types, get rid of prettier.


packages/react-google-maps-api/.eslintrc.js
```diff
@@ -33,8 +34,8 @@ module.exports = {
     'standard',
     'no-inferred-method-name',
     'react-functional-set-state',
-    '@typescript-eslint',
-    'prettier'
+    '@typescript-eslint'
+    // 'prettier'
```
uriklar: I'll open an issue regarding prettier and linting in general because I see we're spending too much of our time over conflicting rules and setups

packages/react-google-maps-api/.eslintrc.js
```diff
@@ -63,6 +64,27 @@ module.exports = {
     }
   },
   rules: {
+    '@getify/proper-arrows/params': [
+      'error', {
+        'unused': 'trailing'
+      }
+    ],
+    '@getify/proper-arrows/name': [
```
uriklar: Is this one really neccesary? or the return one?

> "name": requires => arrow functions to only be used in positions where they receive an inferred name (i.e., assigned to a variable or property, etc), to avoid the poor readbility/debuggability of anonymous function expressions.

I have never seen a codebase that gives names to anonymous arrow functions...

packages/react-google-maps-api/src/GoogleMap.tsx
```diff
@@ -65,101 +65,102 @@ const updaterMap = {
...
+  onZoomChanged?: () => void;
+  onLoad?: (map: google.maps.Map) => void | Promise<void>;
+  onUnmount?: (map: google.maps.Map) => void | Promise<void>;
 }
 
 export class GoogleMap extends React.PureComponent<
   GoogleMapProps,
   GoogleMapState
 > {
-  public static defaultProps: GoogleMapProps = {
```
uriklar: why remove default props?

packages/react-google-maps-api/src/utils/helper.ts
```diff
@@ -6,8 +6,9 @@ const applyUpdaterToNextProps = (
   updaterMap: any,
   prevProps: any,
   nextProps: any,
+  // eslint-disable-next-line @getify/proper-arrows/params
```
uriklar: first of all I think we should remove the params length rule. 
second, easiest way to solve this error is to make this a regular function as it should be

packages/react-google-maps-api/src/components/addons/MarkerClusterer.tsx
```diff
@@ -128,51 +131,72 @@ export class MarkerClustererComponent extends PureComponent<
     markerClusterer: null
   }
 
-  componentDidMount = () => {
+  // eslint-disable-next-line @getify/proper-arrows/this, @getify/proper-arrows/name
+  setMarkerClustererCallback = (): void => {
```
uriklar: We should avoid using class methods as arrow function. 
See: https://medium.com/@rohans16996/dark-side-of-the-fat-arrow-function-8d2a477f743b

TL;DR - when using arrow function as class properties the function get's duplicated on each component instance. Usually this is negligible, but in our lib you might be rendering out thousands of components so this starts to add up

packages/react-google-maps-api/.eslintrc.js
JustFly1984: I found that I can wrap code instead of horizontal scrolling, I can live with it.

packages/react-google-maps-api/.eslintrc.js
JustFly1984: well, I think at some point we will be moving to hooks, and there will be a lot of anonymous functions. I already have all Components refactored to functional in my other project, so I think it is safe to make functional version, and compare bundles on size and performance.

packages/react-google-maps-api/src/utils/helper.ts
JustFly1984: I agree, but I'm not sure how. As I understand, this eslint plugin evolves, and this rules a bit more complex. By now this comments absolutely fine.

packages/react-google-maps-api/src/GoogleMap.tsx
JustFly1984: first of all, if prop is undefined, it is marked as `propName?: string; in interface declaration.

in this case we do not need default props, especially noop functions.

packages/react-google-maps-api/src/components/addons/MarkerClusterer.tsx
JustFly1984: >>>
It`s generally advised to keep *bound* functions out of React Components.
In this case, every single time the component instance is created, a new bound function will be created. which in turn slows down React rendering! Hence, AVOID!

Nothing says about arrow functions in React components, only bound.

JustFly1984: React Components lifecycle methods does not required to be bound, but all other methods are required. Since React async, we are generally avoiding constructors, so the only one way is to keep all instance methods as arrow functions, if we use this in function scope. Otherwise we can just extract that function outside the module to `./src/utils/` or elsewhere.

packages/react-google-maps-api/src/GoogleMap.tsx
uriklar: I agree it's just a matter of taste weather to use defaultProps with noop or a condition checking if it is defined. But just FYI the way Typescript handles default props is that you remove the question mark and the prop is still optional because it has a default. I tested thia behavior and it works. But again, it works this way as well

packages/react-google-maps-api/.vscode/settings.json
```diff
@@ -7,8 +7,8 @@
   "eslint.enable": true,
   "eslint.autoFixOnSave": true,
   "eslint.alwaysShowStatus": true,
-  "eslint.run": "onSave",
-  "eslint.trace.server": "messages",
+  "eslint.run": "onType",
```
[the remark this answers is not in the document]
JustFly1984: it is fine to have "onType" then whole project is typed, and you do new stuff, but I had experienced issues with vscode performance, then I was working on refactoring huge JS codebase to TS, that is why I switched it "onSave".

packages/react-google-maps-api/src/components/addons/MarkerClusterer.tsx
uriklar: Why should we avoid constructors? The only issue about constructor function is that it shouldn't contain side effects. Other than that it's fine.
As you say life cycle methods are automatically bound, but any other function (using this) has to be bound. Using the arrow function syntax compiles to defining the function per instance as the article says

uriklar: Why should we avoid constructors? The only issue about constructor function is that it shouldn't contain side effects. Other than that it's fine.
As you say life cycle methods are automatically bound, but any other function (using this) has to be bound. Using the arrow function syntax compiles to defining the function per instance as the article says

packages/react-google-maps-api/.eslintrc.js
JustFly1984: That is for debuggability - The guy who made this eslint plugin knows what he is doing, but there is no react support out of the box. Maybe we can disable that rule. Well, the thing is that rule asks to refactor () => {} to function functionName() {}

JustFly1984: In general this rule will help less advanced developers to read the code too.

uriklar: Yeah i read the motivation behind the rule 😊 it just feels a bit harsh... I've never seen code that gives names to anonymous function. I have no principal objection to it though

packages/react-google-maps-api/src/GoogleMap.tsx
JustFly1984: I prefer to be consistent) The only one issue which we got, where we could not get rid default props, is functional components, which requires some props default value to be an object or array. If you so default objects or array in props destructuring, each time those props has new object or array, which causing component to re-render. And defaultProps helps with avoiding this.

packages/react-google-maps-api/src/components/addons/MarkerClusterer.tsx
JustFly1984: but bound functions also goes into the object, so there is no gain)

JustFly1984: everything in constructor could be called multiple times, so better to do not run constructor at all in Components. Soon we will migrate to hooks, and there is no constructors.

in react component all functions goes into an object, except if you are extending some other class, which extends Component or PureComponent.

all lifecycle methods are bound by react, all custom function methods should be bound in constructor. bounding a function returns a copy of the function.

Arrow class methods don't have to be bound in babel environment cos we have a proposal babel plugin to handle that case.

packages/react-google-maps-api/src/utils/helper.ts
uriklar:
```
function applyUpdaterToNextProps(
  updaterMap: any,
  prevProps: any,
  nextProps: any,
  instance: any
): any {
  let map: any = {}

  const iter = (fn: any, key: string) => {
    const nextValue = nextProps[key]

    if (nextValue !== prevProps[key]) {
      map[key] = nextValue
      fn(instance, nextValue)
    }
  }

  forEach(updaterMap, iter)

  return map
}
```

this removes the error

packages/react-google-maps-api/src/components/addons/MarkerClusterer.tsx
uriklar: You are absolutely... right! I just read this: https://github.com/facebook/react/issues/9851
Ok so that's ok then.

uriklar:
Issues from testing:
[ ] Docs - Polygon/Polyline/HeatmapLayer examples are broken
[ ] Gatsby - loading script and then opening Data example makes app crash

PR #74, opened 2019-03-31
Use the correct type for mouse handler event

### LINKED ISSUE #73: MouseEvent types are incorrect
`onClick` (and other mouse event handlers) `e` argument type is currently `MouseEvent` but should be `google.maps.MouseEvent`

## Your Environment
@react-google-maps/api version: 1.2.0-alpha.8

## How is it behave?
Currently if you define your own `onClick` handler and check the `e` argument, it is a `MouseEvent` from Google Maps SDK, not a browser `MouseEvent`.

For example with
```jsx
<GoogleMap … onClick={(e) => { console.log(e) }} />
```

You get
```
_.wk {latLng: _.P, va: MouseEvent, pixel: _.M, pa: _.M}
  latLng: _.P {lat: ƒ, lng: ƒ}
  pa: _.M {x: 128.37920009601385, y: 111.05081764442744}
  pixel: _.M {x: 326.203125, y: 316.3125}
  va: MouseEvent {isTrusted: false, screenX: 646, screenY: 636, clientX: 646, clientY: 636, …}
  __proto__: Object
```

Mouse handler events argument type should be `google.maps.MouseEvent`.
This is also true for other components (like `Marker`).

renchap:
This fixes #73

Mouse handler event types are incorrect, at least for `GoogleMap` and `Marker`. They are not browser's `MouseEvent`s but `google.map.MouseEvent`s (see linked issue).

I changed every occurence of it as I guess the problem exists in every component, but only tested it with `GoogleMap` and `Marker`. Please note that `KmlLayer` was already using the correct type.


JustFly1984:
Hello @renchap , sorry, but we do not merge to master yet, only to alpha5 brunch. Also I've already made changes according to your issue, and you can test it in 1.2.0-alpha.11 version

PR #83, opened 2019-04-04
Fix broken initial prop apply + fix some docs examples

uriklar:
# Please explain PR reason.

### DIFF
--- a/packages/react-google-maps-api/src/components/kml/KmlLayer.tsx
+++ b/packages/react-google-maps-api/src/components/kml/KmlLayer.tsx
@@ -67,6 +67,14 @@ export class KmlLayer extends PureComponent<KmlLayerProps, KmlLayerState> {
       this.props.options
     )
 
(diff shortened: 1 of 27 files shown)


packages/react-google-maps-api/src/components/kml/KmlLayer.tsx
```diff
@@ -53,19 +54,24 @@ export class KmlLayer extends PureComponent<KmlLayerProps, KmlLayerState> {
...
+      this.props.onLoad(this.state.kmlLayer)
     }
   }
 
   componentDidMount() {
-    const kmlLayer = new google.maps.KmlLayer(
-      this.props.options
-    )
+    const kmlLayer = new google.maps.KmlLayer({
+      ...this.props.options,
```
JustFly1984: shouldn't it be ```...(this.props.options || {})``` ?

PR #87, opened 2019-04-09
Close InfoWindow on unmount

### LINKED ISSUE #86: infowindows fire twice, one blank
i am trying to load an `InfoWindow` `onMouseOver` of a `Marker`. the `<InfoWindow>` component is a child of `<GoogleMap>`. when I hover over the marker, i get an InfoWindow. when i leave the marker, a smaller, empty infowindow still appears (it loaded behind the first one, which dissapeared `onMouseExit`) They stack up on every hover, and i have to manually "x" them out.

## Your Environment

os: mac
node --version v8.11.4
react version ^16.8.4",
(Created with create-react-app)

@react-google-maps/api version "^1.2.0"

## How is it behave?
two infowindows appear onMouseOver, one is blank.
one infowindow hides onMouseExit.
Screen Shot 2019-04-07 at 10 49 51 PM
Screen Shot 2019-04-07 at 11 03 53 PM
(the BIGGER one is working correctly... the smaller one is the relic...)

one InfoWindow should appear onMouseOver, one should hide onMouseExit

uriklar:
Fixes #86 

The change in only in InfoWindow.tsx but I pushed lib files so we don't forget to when publishing. 
We should add a pre-commit hook


JustFly1984:
there is a `yarn publish` script in package.json, which triggers yarn build before npm publish, so there is no need for pre-hook, and there is no need to commit whole /lib into github.

uriklar:
Ok np I'll fix it in a bit

JustFly1984:
I

PR #89, opened 2019-04-11
Require less props for LoadScript

FredyC:
https://developers.google.com/maps/documentation/javascript/localization

> By default, the Maps JavaScript API uses the user's preferred language setting as specified in the browser,

https://developers.google.com/maps/documentation/javascript/versions

> If you do not explicitly specify a channel or version, you will receive the weekly channel by default.
> For most applications, we recommend the weekly channel. This is the most current and up-to-date version and contains the latest bug fixes and performance improvements.

### DIFF
--- a/packages/react-google-maps-api/src/LoadScript.tsx
+++ b/packages/react-google-maps-api/src/LoadScript.tsx
@@ -13,24 +13,16 @@ interface LoadScriptState {
 
 interface LoadScriptProps {
   // required
-  id: string;
-
-  // required
   googleMapsApiKey: string;
-
-  // required
-  language: string;
-
-  // required
-  region: string;
-
-  // required
+  id: string;
   version: string;
+  language: string;
+  region: string;
+  libraries: string[];
   loadingElement?: React.ReactNode;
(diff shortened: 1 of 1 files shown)


packages/react-google-maps-api/src/LoadScript.tsx
```diff
@@ -40,6 +32,7 @@ const DefaultLoadingElement = () => (
 
 class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
   public static defaultProps = {
+    id: 'script-loader',
     libraries: [] // Do not remove!,
```
uriklar: We should add ```version: "weekly"``` since this is the version used by default

FredyC: I had there for a brief moment, but since it's default within API, I decided to remove it. If someone wants to be explicit, they can specify the version.

FredyC:
How you could you merge this? 😮 It's only a half baked solution since Typescript is still requiring those props. Should not have been doing it in a hurry 🥇

PR #92, opened 2019-04-11
Export useGoogleMap hook

FredyC:
To be able to access map instance in other components for eg. panning needs.


uriklar:
@JustFly1984 @FredyC I believe this PR requires us to bump the react version in our peerDependencies right?

uriklar:
image

FredyC:
Well, not necessarily. It's a kinda gray area. As long as people don't use that hook, they are good with older React. That's why is there that invariant, so they don't use it in the wrong environment.

uriklar:
Right you are! Just validated it in a codesandbox to be sure :-)

PR #93, opened 2019-04-11
Update LoadScriptProps to reflect optionals

FredyC:
Fixed #89 🙈


FredyC:
No, it shouldn't because then you would be opting out from auto-detection based on `Accept-Language`.

uriklar:
@FredyC from my experiments (this is at least true for our setup) when providing ```defaultProps``` you need to remove the ```?``` from the property since it can not be undefined. It will not be required when consuming the component.

PR #97, opened 2019-04-12
Fix LoadScript defaultProps and interface defenitions

uriklar:
see: https://github.com/JustFly1984/react-google-maps-api/pull/93#issuecomment-482439401

### DIFF
--- a/packages/react-google-maps-api/src/LoadScript.tsx
+++ b/packages/react-google-maps-api/src/LoadScript.tsx
@@ -14,8 +14,8 @@ interface LoadScriptState {
 export interface LoadScriptProps {
   // required
   googleMapsApiKey: string;
-  id?: string;
-  version?: string;
+  id: string;
+  version: string;
   language?: string;
   region?: string;
   libraries?: string[];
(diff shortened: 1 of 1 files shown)


FredyC:
I don't think it's needed. Check the https://codesandbox.io/s/l9z4ropo2m I made. Not sure if you have something in your setup that would require that in here.

uriklar:
It's causing ts errors. Take this line for example:
https://github.com/JustFly1984/react-google-maps-api/blob/1f8d47e029f754fd8fc903cdc86b7e8d9acf39eb/packages/react-google-maps-api/src/LoadScript.tsx#L137

When id is nullable it gives an error even though a default prop is defined. Sounds like something typescript should know how to handle but for some reason it doesn't

uriklar:
Checkout: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#support-for-defaultprops-in-jsx

In their example they removed the question mark

FredyC:
Ok, I don't mind that much. Class components suck :) I think that eventually, it would be great to rewrite internals to functional components. It would make a lot of things easier. But of course it would become incompatible with older React and I suppose that's something your don't really want.

uriklar:
We're definitely meaning to do the hooks re-write. It's already WIP. That way all (or almost all) components can use a simple ```useMapComponent ``` hook instead of all this duplicated lifecycle method logic.

We're already demanding 16.6.0 as a peerDep so I don't see any problem with demanding 16.8.0. It's an easy migration...

PR #98, opened 2019-04-12
Fixed duplicate react types issue 

### LINKED ISSUE #96: Duplicate identifier 'LibraryManagedAttributes'.
**Version: 1.2.1' or '1.2.0'**

I got this error 'Duplicate identifier 'LibraryManagedAttributes'. when I run the application after installing 1.2.1 package or 1.2.0 package. 

> C:/...web/node_modules/@react-google-maps/api/node_modules/@types/react/index.d.ts
> (2777,14): Duplicate identifier 'LibraryManagedAttributes'.

seems like there is an issue with type/react version issue. How do I solve this?

Similiar issue?: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33822

- OS: Windows
- Node: v10.15.1
- "react": "^16.8.4",
- "react-bootstrap": "^1.0.0-beta.6",
- "react-dom": "^16.8.4",
- "react-router": "^5.0.0",
- "react-router-dom": "^5.0.0",

uriklar:
Fixes #96

### DIFF
--- a/packages/react-google-maps-api/package.json
+++ b/packages/react-google-maps-api/package.json
@@ -78,18 +78,16 @@
   "dependencies": {
     "@react-google-maps/marker-clusterer": "1.2.2",
     "@types/googlemaps": "3.30.19",
-    "@types/invariant": "2.2.29",
     "invariant": "2.2.4"
   },
   "peerDependencies": {
-    "@types/react": "^16.8.10",
-    "@types/react-dom": "^16.8.3",
     "react": "^16.6.3",
     "react-dom": "^16.6.3"
   },
   "devDependencies": {
     "@getify/eslint-plugin-proper-arrows": "7.1.0",
     "@types/babel-types": "7.0.6",
+    "@types/invariant": "2.2.29",
     "@types/jest": "^24.0.11",
     "@types/react": "^16.8.10",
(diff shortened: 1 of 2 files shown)


FredyC:
This is very tricky and I fighting with it every day. It's an annoying issue with Yarn. Anytime you update `@types/react`, other packages relying on it (even with `*` as version) will keep the old version. And that creates duplicates. The only cure I found so far is manually to edit `yarn.lock` and remove all occurrences and re-run.

However, in your case, your packages are requiring fixed types package version which makes it even worse.

I would recommend switching to Yarn workspaces which are [supported with Lerna](https://github.com/lerna/lerna/tree/master/commands/bootstrap#--use-workspaces) which leads to a single `yarn.lock` and it's much easier to tackle this problem there.

uriklar:
I removed the @types packages from being peerDependencies. I think this should resolve the issue. Deleting yarn.lock and rerunning currently fixes the issue for me

FredyC:
Ok, but still, Yarn workspaces are much better. It's also easier for contributors as they don't need to bother with Lerna at all and just `yarn install` in a root.

uriklar:
I'll check it out. Personally, i'd get rid of lerna altogether and have just one main package

PR #109, opened 2019-04-29
fix(api): Redraw OverlayView on position or bounds change

FredyC:
https://spectrum.chat/react-google-maps/general/overlayview-position-is-not-updated~3a32f12a-b909-4316-af7a-cd6f9a92ed8c

Unfortunately, I was not able to test it out, build is failing for me with bunch of TS errors.

### DIFF
--- a/packages/react-google-maps-api/src/components/dom/OverlayView.tsx
+++ b/packages/react-google-maps-api/src/components/dom/OverlayView.tsx
@@ -140,6 +140,12 @@ export class OverlayView extends React.PureComponent<
 
     this.setState(setOverlayView, this.setOverlayViewCallback)
   }
+  
+  componentDidUpdate(prevProps: OverlayViewProps) {
+    if (prevProps.position !== this.props.position || prevProps.bounds !== this.props.bounds) {
+      setTimeout(() => this.state.overlayView!.draw(), 0);
+    }
+  }
 
   componentWillUnmount() {
     if (this.state.overlayView !== null) {


packages/react-google-maps-api/src/components/dom/OverlayView.tsx
```diff
@@ -140,6 +140,12 @@ export class OverlayView extends React.PureComponent<
 
     this.setState(setOverlayView, this.setOverlayViewCallback)
   }
+  
+  componentDidUpdate(prevProps: OverlayViewProps) {
```
uriklar: @FredyC I guess we need to add ```getPixelPositionOffset``` to this check no? 
Also, why the setTimeout? I see the original lib does it too but not sure why...

FredyC: Yea, I did it based on original considering it was working. Hard to say what is the reasoning behind it.

uriklar: I can only guess he didn't want it to block rendering..  Though, we're not really rendering anything in the component besides it's children. I'll need to test it out but it feels like this component could have

```
shouldComponentUpdate() {
   return false;
}
```

Or maybe just re-render only if the children prop changed

FredyC: I don't think you can compare children prop, it's always going to be a different object unless someone would actually `React.useMemo` it :) Also `getPixelPositionOffset` might in theory be dynamic. I wouldn't over optimize here, the `PureComponent` should suffice or later `React.memo` if we get rid of the class.

PR #121, opened 2019-05-08
[OverlayView] Add check for containerElement

FredyC:
As much as awkward this looks, it seems that on some occasion the `draw` is called before the `onAdd` causing the error. It seems to be able to recover itself with a next draw, but it's kinda spamming our bug reporting service.

> Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'

image


JustFly1984:
published alpha-10

FredyC:
Thanks, it will probably show over time if it helped. I wasn't able to actually reproduce it yet.

PR #133, opened 2019-05-27
Added two new mouse Events to help on mobile devices

herzuull:
Added 2 new event from the native events list of Google maps.
Those 2 new events add the opportunity to the user to listen at the mouse down and mouse up events.
Can help to create a method to manage long click on mobile devices.


javiercr:
Shouldn't these two events be added to the `GoogleMapProps` as well?

PR #136, opened 2019-06-02
Smooth center

### LINKED ISSUE #134: Add smooth center change
How about adding `smoothPan` (`boolean`) prop for GoogleMap component? When value is `true` `center` prop change uses smooth map.panTo method to move the camera (https://developers.google.com/maps/documentation/javascript/reference/map#Map.panTo). 
I could make the PR if it is sounds interesting.

regulyarniy:
Add `smoothCenter` prop in `GoogleMap`  for smooth camera fly on change.
Fixes #134

### DIFF
--- a/packages/react-google-maps-api/src/GoogleMap.tsx
+++ b/packages/react-google-maps-api/src/GoogleMap.tsx
@@ -43,7 +43,7 @@ const updaterMap = {
   ) {
     map.setCenter(center)
   },
-  smoothCenter(
+  panTo(
     map: google.maps.Map,
     center: google.maps.LatLng | google.maps.LatLngLiteral
   ) {
@@ -110,7 +110,7 @@ export interface GoogleMapProps {
   onZoomChanged?: () => void;
   onLoad?: (map: google.maps.Map) => void | Promise<void>;
   onUnmount?: (map: google.maps.Map) => void | Promise<void>;
-  smoothCenter?: google.maps.LatLng | google.maps.LatLngLiteral;
+  panTo?: google.maps.LatLng | google.maps.LatLngLiteral;
 }
 
 export class GoogleMap extends React.PureComponent<GoogleMapProps, GoogleMapState> {
--- a/packages/react-google-maps-api/src/GoogleMap.tsx
+++ b/packages/react-google-maps-api/src/GoogleMap.tsx
@@ -43,6 +43,12 @@ const updaterMap = {
   ) {
     map.setCenter(center)
   },
+  smoothCenter(
+    map: google.maps.Map,
+    center: google.maps.LatLng | google.maps.LatLngLiteral
+  ) {
+    map.panTo(center)
+  },
   clickableIcons(map: google.maps.Map, clickable: boolean) {
     map.setClickableIcons(clickable)
   },
@@ -104,6 +110,7 @@ export interface GoogleMapProps {
   onZoomChanged?: () => void;
   onLoad?: (map: google.maps.Map) => void | Promise<void>;
   onUnmount?: (map: google.maps.Map) => void | Promise<void>;
+  smoothCenter?: google.maps.LatLng | google.maps.LatLngLiteral;
 }
 
 export class GoogleMap extends React.PureComponent<GoogleMapProps, GoogleMapState> {


regulyarniy:
I had issues with storybook deps not updated with v1.2.3 published, so the first commit fixes it.

xalisys:
@regulyarniy , by looking at the changes doesn't seems to be necessary to introduce this api to GoogleMap.tsx , you can simple do something similar, in you code.

`
const map=useGoogleMap(); 
map.setCenter({.....});
`

agree with @uriklar , keeping the api cleaner.

regulyarniy:
@uriklar good idea! I renamed it to `panTo`
@xalisys Of cource it could be done with map instance, but the user should think about how to wrap this prop in parent component(or use hook as your example) - it's not obvious for everyone. this prop exposes basic google api to the component like any other props (`zoom`, `center` etc..) - why you think it's not necessary? I spent about an hour to figure out how to make a smooth camera transition in my project. I think this prop will help others not to spend so much time on this feature, and in my opinion it does not contradict the purpose of the component at all - to reflect the google api from map instance to React props

JustFly1984:
I think we can add this to docs and/or examples and codesandbox, and link it to change log.

packages/react-google-maps-api/src/GoogleMap.tsx
```diff
@@ -104,6 +110,7 @@ export interface GoogleMapProps {
   onZoomChanged?: () => void;
   onLoad?: (map: google.maps.Map) => void | Promise<void>;
   onUnmount?: (map: google.maps.Map) => void | Promise<void>;
+  panTo?: google.maps.LatLng | google.maps.LatLngLiteral;
```
JustFly1984: Shouldn't `panTo` be a function type, accepting an argument and returning `void`?

uriklar: Maybe the name is confusing. The idea is to pass a latlang, that whenever it changes that map will panTo it. I thought using the panTo name would make it clear that this uses the panTo functionality.

JustFly1984: the name is confusing.

JustFly1984:
@ regulyarniy Can you please move storybook fixtures to separate PR so I could merge it separately?

regulyarniy:
> @ regulyarniy Can you please move storybook fixtures to separate PR so I could merge it separately?

@JustFly1984 I dropped the commit. Thanks

packages/react-google-maps-api/src/GoogleMap.tsx
regulyarniy: Maybe `panToCenter` will be less confusing and sounds more meaningful in a list with `center` prop?

JustFly1984: maybe rename panTo to autopanCoordinates ?

regulyarniy: I assume that user will prefer to use only one type of center change - instant (`center`) or panned (`???newPropName???`) but using both in same time could potentially cause to undesired effects - because methods `setCenter` and `panTo` called if only prop changed. I must insist to use word `center` as suffix or prefix of propname to highlight its potential problems with `center` prop used simultaneously. Or we could throw exception if user uses both, but I could imagine the case when both could be useful.

example: https://codesandbox.io/s/markers-playbox-rodck - top and bottom markers using `panTo` methods, middle uses `setCenter`. try click top, middle, bottom, middle, top - second middle click won't move the camera because its not changed `center` prop passed to component. it's could be confusing when use both props.

What are your thoughts, guys?

JustFly1984: @uriklar mention

JustFly1984: I'm pretty much settled to do not expand an API, and to do not overcomplicate things. Can't say yes for every feature, sorry. I feel that it could bite performance in general, especially for people who do not use eslint-plugin-react-perf. Providing objects as props to components is not good idea in general, and we are providing objects cos `react-google-maps` requires to. I feel that it is better to leave it as it is, and use ```const map=useGoogleMap(); map.setCenter({.....});``` as @xalisys  proposed, without drawbacks to performance for everybody. Everybody who has same requirements will google up this issue and use it.

regulyarniy: @JustFly1984 
couple of questions:
1.  So you are planning to remove\refactor `center` prop in future because it could affect performance too? It have analogous problem. With all the respect to you "performance problem" sounds for me like the non-logic excuse.
2. Where `useGoogleMap`  should be used? Docs have not clear example. It exposes map instance to map children throught context? How to use it in parent component?

TrejGun: @JustFly1984 Yeah it is a good question, how can i get center in map events?

```js
<GoogleMap
    onBoundsChanged={() => {
        // no reference to `map` or` center` is available
    }}
  >
    {...}
</GoogleMap>
```

PR #141, opened 2019-06-05
Use TSDX building tool

FredyC:
In reaction to broken 1.2.3 build, I first wanted to simply fix that by including `umd` folder. However, then I realized it's an overly complicated way to build things and since I lately had a great experience using [TSDX](https://www.npmjs.com/package/tsdx) tool so I went a bit further and used it here as well.

The main difference is that instead of multiple folders with similar content, there is a `dist` folder that contains these files (plus a bunch of .d.ts files). The `index.js` serves as the main entry point which picks either development or production bundle based on NODE_ENV. The difference, in this case, is 40kB, so that's a nice win.

```
index.d.ts
index.js
reactgooglemapsapi.cjs.development.js
reactgooglemapsapi.cjs.development.js.map
reactgooglemapsapi.cjs.production.js
reactgooglemapsapi.cjs.production.js.map
reactgooglemapsapi.es.production.js
reactgooglemapsapi.es.production.js.map
reactgooglemapsapi.umd.development.js
reactgooglemapsapi.umd.development.js.map
reactgooglemapsapi.umd.production.js
reactgooglemapsapi.umd.production.js.ma
```


uriklar:
+1 on this one too :-)
I saw this tool and thought it would be great for us. I feel our current TS setup is overly complicated and not optimal..
I would be happy to pass some of this complexity on to a tool that already handles it.

packages/react-google-maps-api/src/components/dom/OverlayView.tsx
```diff
@@ -171,8 +171,8 @@ export class OverlayView extends React.PureComponent<
         this.containerElement
       )
     ) : (
-      <></>
-    )
+        <></>
```
uriklar: any idea why these extra spaces are being added? They give me a linting error

FredyC: Well, that something I wanted to tweak later as well. I think linter should not be used for styling, we are living in the age of Prettier. I could revert that from this PR if it bothers you too much though.

FredyC: Done. I've also added `prepublishOnly` script now to avoid the mess that happened earlier on Spectrum :)

uriklar: I am very much pro prettier. I've done a ton of work to get to the point where we are now. Before that me and @JustFly1984 kept having conflicting configurations.. 
Right now we're using prettier, but via eslint. And using eslint's autofix to format the code. Not sure why it doesn't work correctly in your env.

FredyC: I have a Prettier extension and it was reformatting every file I've touched, so I disabled it. Are you aware that you have Prettier disabled in Eslint config? Doesn't matter now, let's revisit that later.

package.json
```diff
@@ -28,6 +28,7 @@
     "test:api": "cd packages/react-google-maps-api && yarn test && cd ../.. ",
     "bootstrap": "lerna bootstrap",
     "build": "lerna run build --stream",
+    "clean": "rimraf ./yarn.lock ./packaje-lock.json ./node_modules/ && yarn",
```
FredyC: `packaje-lock` doesn't look good :D

Issue #137, opened 2019-06-03
add support for client parameter


modosc:
google supports client keys instead of api keys:

https://developers.google.com/maps/documentation/javascript/get-api-key#specifying-a-client-id-when-loading-the-api

## How is it behave?
we don't have an api key so LoadScript doesn't work correctly. there's no way to provide a client key instead.

## How should it behave correctly?
client key should be supported

uriklar:
Thanks for the issue! Wanna try creating a PR?

PR #143, opened 2019-06-06
Add useLoadScript and LoadScriptNext

### LINKED ISSUE #139: LoadScript used multiple times in app
I've already tried to discuss this [on the spectrum](https://spectrum.chat/react-google-maps/general/you-have-included-the-google-maps-javascript-api-multiple-times-on-this-page~99cc4656-9502-4fc4-9fa2-9cdccc8f44f7), but no response, so let's poke a bear in here :)

There is an obvious problem that `LoadScript` injects a script to header every time it is mounted. It's not keeping any information it was already loaded before. This is not only a problem of fetching the same thing multiple times, but Google API seems to have a problem with that and it causes weird issues like __`google` is not defined__

Furthermore, it shouldn't be just a boolean flag to remember it has been loaded, because it might be needed with a different set of props.

The react-google-maps are using `scriptjs` module and I have a feeling it actually tracks fetching the same resource multiple times and that's why it works.

So why you have opted not to use `scriptjs` here I wonder?

FredyC:
Fixes #139 

As noted in #139, instead of modifying current LoadScript I have created an alternative approach with Hooks and without that cleanup logic which I think was causing more issues than it helped.

I've added this build into our app and so far it works correctly and intend to push it to the production soon (yes, I am that bold) so we will see if it actually helps.

I've also decided to keep whole logic in the `useLoadScript` hook instead of a component to provide loading state directly instead of the callback. This is the way I am using it and works great. 

```tsx
export const BaseMap = observer<IBaseMapProps>(
  ({ model, children }) => {
    const { user } = useRoot()

    const { isLoaded } = useLoadScript({
      id: 'script-loader',
      version: 'weekly',
      googleMapsApiKey: appConfig.googleMapsApi!,
      libraries,
      language: user.language,
    })

    const renderMap = () => (
      <GoogleMap
        zoom={model.zoom}
        center={model.center}
        onCenterChanged={model.updatePosition}
        onZoomChanged={model.updateZoom}
        onClick={model.onClick}
        onLoad={model.setMapRef}
        onUnmount={() => model.setMapRef(null)}
        options={getMapOptions(isMobile)}
      >
        {children}
      </GoogleMap>
    )

    return isLoaded ? renderMap() : <Loading />
  },
)
```

For the convenience (and possible migration path), there is the `LoadScriptNext` component, but personally, I think it's not that useful anymore.

cc @JustFly1984 @uriklar @orta-sanz

### DIFF
--- a/packages/react-google-maps-api/src/LoadScript.tsx
+++ b/packages/react-google-maps-api/src/LoadScript.tsx
@@ -5,6 +5,7 @@ import { preventGoogleFonts } from "./utils/prevent-google-fonts"
 
 import { isBrowser } from "./utils/isbrowser"
 import { LoadScriptUrlOptions, makeLoadScriptUrl } from "./utils/make-load-script-url";
+import invariant from "invariant";
 
 let cleaningUp = false
 
@@ -13,6 +14,7 @@ interface LoadScriptState {
 }
 
 export interface LoadScriptProps extends LoadScriptUrlOptions {
+  id?: string;
   loadingElement?: React.ReactNode;
   onLoad?: () => void;
   onError?: (error: Error) => void;
@@ -173,6 +175,8 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
     if (this.props.preventGoogleFontsLoading) {
       preventGoogleFonts()
     }
+
+    invariant(typeof this.props.id === 'string', 'LoadScript requires "id" prop to be a string')
 
     const injectScriptOptions = {
       id: this.props.id!,
--- a/packages/react-google-maps-api/src/LoadScript.tsx
+++ b/packages/react-google-maps-api/src/LoadScript.tsx
@@ -165,7 +165,7 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
     Array.prototype.slice
       .call(document.getElementsByTagName("style"))
       .filter(function filter(style: HTMLStyleElement): boolean {
-        return style.innerText && style.innerText.includes(".gm-")
+        return style.innerText.length > 0 && style.innerText.includes(".gm-")
       })
       .forEach(function forEach(style: HTMLStyleElement) {
         if (style.parentNode) {
--- a/packages/react-google-maps-api/src/LoadScript.tsx
+++ b/packages/react-google-maps-api/src/LoadScript.tsx
@@ -26,15 +26,17 @@ export interface LoadScriptProps {
   preventGoogleFontsLoading?: boolean;
 }
 
-const DefaultLoadingElement = () => (
-  <div>{`Loading...`}</div>
-)
+export function DefaultLoadingElement() {
+  return <div>{`Loading...`}</div>
+}
+
+export const defaultLoadScriptProps = {
+  id: 'script-loader',
+  version: 'weekly'
+}
 
 class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
-  public static defaultProps = {
-    id: 'script-loader',
-    version: 'weekly',
-  }
+  public static defaultProps = defaultLoadScriptProps
 
   check: React.RefObject<HTMLDivElement> = React.createRef()
 
@@ -108,7 +110,7 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
 
   // eslint-disable-next-line @getify/proper-arrows/name
   isCleaningUp = async () => {
-    function promiseCallback (resolve: () => void) {
+    function promiseCallback(resolve: () => void) {
       if (!cleaningUp) {
         resolve()
       } else {
@@ -140,10 +142,10 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
 
     Array.prototype.slice
       .call(document.getElementsByTagName("script"))
-      .filter(function filter (script: HTMLScriptElement) {
+      .filter(function filter(script: HTMLScriptElement) {
         return script.src.includes("maps.googleapis")
       })
-      .forEach(function forEach (script: HTMLScriptElement) {
+      .forEach(function forEach(script: HTMLScriptElement) {
         if (script.parentNode) {
           script.parentNode.removeChild(script)
         }
@@ -151,11 +153,11 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
 
     Array.prototype.slice
       .call(document.getElementsByTagName("link"))
-      .filter(function filter (link: HTMLLinkElement) {
+      .filter(function filter(link: HTMLLinkElement) {
         link.href ===
-        "https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Google+Sans"
+          "https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Google+Sans"
       })
-      .forEach(function forEach (link: HTMLLinkElement) {
+      .forEach(function forEach(link: HTMLLinkElement) {
         if (link.parentNode) {
           link.parentNode.removeChild(link)
         }
@@ -163,10 +165,10 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
 
     Array.prototype.slice
       .call(document.getElementsByTagName("style"))
-      .filter(function filter (style: HTMLStyleElement) {
+      .filter(function filter(style: HTMLStyleElement) {
         return style.innerText && style.innerText.includes(".gm-")
       })
-      .forEach(function forEach (style: HTMLStyleElement) {
+      .forEach(function forEach(style: HTMLStyleElement) {
         if (style.parentNode) {
           style.parentNode.removeChild(style)
         }
@@ -209,7 +211,7 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
           this.props.onLoad()
         }
 
-        this.setState(function setLoaded () {
+        this.setState(function setLoaded() {
           return {
             loaded: true
           }
--- a/packages/react-google-maps-api/src/LoadScript.tsx
+++ b/packages/react-google-maps-api/src/LoadScript.tsx
@@ -4,6 +4,7 @@ import { injectScript } from "./utils/injectscript"
 import { preventGoogleFonts } from "./utils/prevent-google-fonts"
 
 import { isBrowser } from "./utils/isbrowser"
+import invariant from "invariant";
 
 let cleaningUp = false
 
@@ -12,10 +13,10 @@ interface LoadScriptState {
 }
 
 export interface LoadScriptProps {
-  // required
-  googleMapsApiKey: string;
-  id: string;
-  version: string;
+  googleMapsApiKey?: string;
+  googleMapsClientId?: string;
+  id?: string;
+  version?: string;
   language?: string;
   region?: string;
   libraries?: string[];
@@ -134,7 +135,7 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
 
   cleanup = () => {
     cleaningUp = true
-    const script = document.getElementById(this.props.id)
+    const script = document.getElementById(this.props.id!)
 
     if (script && script.parentNode) {
       script.parentNode.removeChild(script)
@@ -176,30 +177,40 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
 
   // eslint-disable-next-line @getify/proper-arrows/this, @getify/proper-arrows/name
   injectScript = () => {
-    if (this.props.preventGoogleFontsLoading) {
+    const { googleMapsApiKey, googleMapsClientId, id, version, language, region, libraries, preventGoogleFontsLoading } = this.props
+
+    if (preventGoogleFontsLoading) {
       preventGoogleFonts()
     }
 
-    const params = [`key=${this.props.googleMapsApiKey}`]
+    const params = []
 
-    if (this.props.version) {
-      params.push(`v=${this.props.version}`)
+    if (googleMapsApiKey) {
+      params.push(`key=${googleMapsApiKey}`)
+    } else if (googleMapsClientId) {
+      params.push(`client=${googleMapsClientId}`)
+    } else {
+      invariant(false, "You need to specify either googleMapsApiKey or googleMapsClientId for @react-google-maps/api load script to work.")
     }
 
-    if (this.props.language) {
-      params.push(`language=${this.props.language}`)
+    if (version) {
+      params.push(`v=${version}`)
     }
 
-    if (this.props.region) {
-      params.push(`region=${this.props.region}`)
+    if (language) {
+      params.push(`language=${language}`)
     }
 
-    if (this.props.libraries && this.props.libraries.length) {
-      params.push(`&libraries=${this.props.libraries.join(",")}`)
+    if (region) {
+      params.push(`region=${region}`)
+    }
+
+    if (libraries && libraries.length) {
+      params.push(`&libraries=${libraries.join(",")}`)
     }
 
     const injectScriptOptions = {
-      id: this.props.id,
+      id: id!,
       url: `https://maps.googleapis.com/maps/api/js?${params.join('&')}`
     }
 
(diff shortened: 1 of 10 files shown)


packages/react-google-maps-api/src/useLoadScript.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+        console.warn(`
+        There has been an Error with loading Google Maps API script, please check that you provided correct google API key to <LoadScript /> (${googleMapsApiKey})
+        Otherwise it is a Network issues.
+      `)
+        console.error(err)
+      })
+
+    return () => {
+      setLoadedIfMounted(false)
```
FredyC: I am thinking this might be actually a bad idea. It would be causing a switch back to the loading state on unmount and possibly some unwanted flicker.

packages/react-google-maps-api/src/useLoadScript.ts
```diff
@@ -0,0 +1,117 @@
...
+    return () => {
+      isMounted.current = false
+      if (onUnmount) {
+        onUnmount()
+      }
+    }
+  }, [])
+
+  React.useEffect(() => {
+    if (isBrowser && preventGoogleFontsLoading) {
```
FredyC: I haven't fully understood why is this needed, so I kept it, but `react-google-maps` had nothing similar so I wonder what's the purpose of it?

JustFly1984: That is my own addition, cos I do not want to google maps to include google font Roboto. I already have Roboto fonts loaded with async in my app.

packages/react-google-maps-api/src/useLoadScript.ts
JustFly1984: that effect is only running if id or url has changed

JustFly1984:
Where did you import that `observer` function from?

FredyC:
@JustFly1984 Nevermind, that's just MobX stuff, doesn't matter for the sake of the example :)

packages/react-google-maps-api/src/useLoadScript.ts
```diff
@@ -0,0 +1,117 @@
...
+        There has been an Error with loading Google Maps API script, please check that you provided correct google API key to <LoadScript /> (${googleMapsApiKey})
+        Otherwise it is a Network issues.
+      `)
+        console.error(err)
+      })
+
+    return () => {
+      setLoadedIfMounted(false)
+    }
+  }, [id, url])
```
FredyC: I should probably include the `onLoad` and `onError` here as well, but it's a kinda bad idea to be re-injecting script if a parent is not having `useCallback` for these. So I think I will have to keep those in a ref otherwise it would see only first instances. Any idea of how to tackle it better?

packages/react-google-maps-api/src/useLoadScript.ts
FredyC: Yea, but also on unmount of the component which would cause a short flicker back to a loading state.

packages/react-google-maps-api/src/useLoadScript.ts
FredyC: Nevermind, figured out a better way with separate useEffect calls. Although it might be better to keep those callbacks out of the hook and use them only for a component variant.

JustFly1984:
I think in example you should not generate `getMapOptions(isMobile)` on each function call, but use ```React.memo(getMapOptions(isMobile), [isMobile])```

FredyC:
> I think in example you should not generate `getMapOptions(isMobile)` on each function call, but use `React.memo(getMapOptions(isMobile), [isMobile])`

That's premature optimization as is a plain object. Have a look at the article from Kent I've seen yesterday https://kentcdodds.com/blog/usememo-and-usecallback

Please, let's focus on the PR itself, I was just lazy to strip down that example more, wanted to show how that hook can be used :)

JustFly1984:
useLoadScript should accept second argument `[appConfig.googleMapsApi, user.language]`

I have specific case where I want maps language to switch to correct language, when user changes the language in interface. This autodetection language feature is good enough, until you sit in some hotel in vietnam and try to use maps in vietnamese, cos it detects your browser. :-/

FredyC:
> useLoadScript should accept second argument `[appConfig.googleMapsApi, user.language]`

I don't follow, why would be a benefit of the second argument? With component, it's also part of single props object.

packages/react-google-maps-api/src/LoadScriptNext.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { LoadScriptProps, DefaultLoadingElement } from './LoadScript'
+import { useLoadScript } from './useLoadScript'
+
+export interface LoadScriptNextProps extends LoadScriptProps {
+  loadingElement: React.ReactElement
+  children: React.ReactElement
+}
+
+export const LoadScriptNext = React.memo<LoadScriptNextProps>(
+  ({ loadingElement = <DefaultLoadingElement />, children, ...props }) => {
```
JustFly1984: Please don't use default values for props on destructuring and use defaultProps instead. performance requirements.

packages/react-google-maps-api/src/LoadScriptNext.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { LoadScriptProps, DefaultLoadingElement } from './LoadScript'
+import { useLoadScript } from './useLoadScript'
+
+export interface LoadScriptNextProps extends LoadScriptProps {
+  loadingElement: React.ReactElement
+  children: React.ReactElement
+}
+
+export const LoadScriptNext = React.memo<LoadScriptNextProps>(
```
JustFly1984: Please declare a component before wrapping it to React.memo. performance reasons.

FredyC: Huh? Never head that one before. How is more performant if it's separated?

FredyC:
> Please use named functions instead of arrow functions. better for debug, and `function` declarations hoist.

I already did that. Or did I missed a spot? :)

packages/react-google-maps-api/src/LoadScriptNext.tsx
JustFly1984: The correct way to declare a component as function first:
```jsx
function MyComponent () {
  return <JSX />
}
```

then you export it wrapped in memo:
```jsx
export default React.memo(MyComponent)
```

otherwise it ruins React debug with <undefined /> components.

JustFly1984: React does check for Component.displayName, and React.memo does not provide it for us.

FredyC: So you mean "debugging reasons". You got me fairly confused how that can be related to performance 😆

packages/react-google-maps-api/src/LoadScriptNext.tsx
FredyC: How do you handle that with TypeScript? I mean if I do `function LoadScriptNext` it does not have `defaultProps` and I would have to `ts-ignore` it which is ugly. The only way I know is to keep `const` which will keep the name as well, but allow proper typing.

```tsx
const LoadScriptNext: React.FC<LoadScriptNextProps> = ({ loadingElement, children, ...props }) => {
  const { isLoaded } = useLoadScript(props)
  return isLoaded ? children : loadingElement
}

LoadScriptNext.defaultProps = {
  loadingElement: <DefaultLoadingElement />
}

export default React.memo(LoadScriptNext)

```

FredyC: Another approach would be like that, but I think const typing is much easier and approachable.

```ts
const defaultLoadingElement = <DefaultLoadingElement />

function LoadScriptNext({ loadingElement, children, ...props }: LoadScriptNextProps) {
  const { isLoaded } = useLoadScript(props)
  return isLoaded ? children : loadingElement || defaultLoadingElement
}

export default React.memo(LoadScriptNext)
```

packages/react-google-maps-api/src/LoadScriptNext.tsx
JustFly1984: sorry)

packages/react-google-maps-api/src/LoadScriptNext.tsx
JustFly1984:
```jsx
function LoadScriptNext(props: LoadScriptNextProps) {
  const { isLoaded } = useLoadScript(props)

  return isLoaded ? props.children : props.loadingElement
}

LoadScriptNext.defaultProps = {
  loadingElement: <DefaultLoadingElement />
}

export default React.memo(LoadScriptNext)
```

packages/react-google-maps-api/src/LoadScript.tsx
```diff
@@ -26,15 +26,15 @@ export interface LoadScriptProps {
   preventGoogleFontsLoading?: boolean;
 }
 
-const DefaultLoadingElement = () => (
-  <div>{`Loading...`}</div>
-)
+export const DefaultLoadingElement = () => <div>{`Loading...`}</div>
```
JustFly1984: () =>

packages/react-google-maps-api/src/useLoadScript.ts
```diff
@@ -0,0 +1,132 @@
+import * as React from 'react'
+
+import { injectScript } from './utils/injectscript'
+import { preventGoogleFonts } from './utils/prevent-google-fonts'
+
+import { isBrowser } from './utils/isbrowser'
+import { LoadScriptProps, defaultLoadScriptProps } from './LoadScript'
+
+export const useLoadScript = ({
```
JustFly1984: () =>

FredyC: Ah, this is the shame. Babel is actually handling that more properly and transpiles to named function. Sadly TypeScript doesn't bother, so yea, I will change that as well.

FredyC:
Ok, I did a couple of changes, hopefully, it's better now :)

packages/react-google-maps-api/src/LoadScriptNext.tsx
FredyC: And what's the benefit of using `defaultProps` really? It did make sense for class components, but here it feels rather obsolete.

JustFly1984: defaultProps does not get redeclared in compare with destructuring default values

JustFly1984: basic types is pretty much safe, but objects, arrays and functions as default values recreated on each function call :-/

FredyC: Sure, if you do `const defaultLoadingElement = <DefaultLoadingElement />` and referenced that, it's essentially the same thing, just bit more manual.

FredyC:
@JustFly1984 Why that last commit? TypeScript will infer that return type on its own, doesn't it?

packages/react-google-maps-api/src/LoadScriptNext.tsx
JustFly1984: Component.defaultProps is not deprecated, and it is a recommended way to set default props :-)

FredyC: Recommended? Would love to see docs link explaining that? :) 

image

FredyC:
I just noticed open PR #137 and included that change in this PR as well.

FredyC:
Rebased to master and ready for review :)

packages/react-google-maps-api/src/useLoadScript.ts
```diff
@@ -0,0 +1,124 @@
...
+    prevLibraries.current = libraries
+  }, [libraries])
+
+  function makeUrl() {
+    const params = []
+
+    if (googleMapsApiKey) {
+      params.push(`key=${googleMapsApiKey}`)
+    } else if (googleMapsClientId) {
+      params.push(`client=${googleMapsClientId}`)
```
JustFly1984: if there is googleMapsClientId, no need to set googleMapsApiKey, and overwise, so basically googleMapsClientId should be first in if/else statement.

packages/react-google-maps-api/src/useLoadScript.ts
```diff
@@ -0,0 +1,124 @@
...
+import { isBrowser } from './utils/isbrowser'
+import { defaultLoadScriptProps } from './LoadScript'
+import invariant from 'invariant';
+
+export interface UseLoadScriptOptions {
+  googleMapsApiKey?: string;
+  googleMapsClientId?: string;
+  id?: string;
+  version?: string;
+  language?: string;
```
JustFly1984: How do I change language in runtime?

packages/react-google-maps-api/src/useLoadScript.ts
FredyC: I don't follow why the order matters, it's either API key or Client ID.

packages/react-google-maps-api/src/useLoadScript.ts
FredyC: Just pass a new value to the hook? It will assemble new url and effect will re-rerun.

packages/react-google-maps-api/src/useLoadScript.ts
JustFly1984: well, both options is optional, but if both options will be entered, clientId should win

FredyC: In the docs I only found this statement ...

> Note: If you were previously using an API key for authentication and are switching to using a client ID, you must remove the key parameter before loading the API. The API will fail to load if both a client ID and an API key are included.

It doesn't explicitly say that "client" is somehow "better". I think it should actually throw invariant to prevent surprising behavior. Someone might expect that key would be winning here. I will tweak that tomorrow.

packages/react-google-maps-api/src/useLoadScript.ts
FredyC: Btw, in reality, changing language is not something the user does very often. For a fully localized app, it's usually better to just do a full reload to make sure everything is correctly loaded. It's surely a cool effect when it's changed in runtime, but not worth the hassle imo.

JustFly1984: in reality this is a part of my own app functionality, and it is important for my own project. sadly google doesn't provide an easy api to switch locale in runtime at all. I had this issue with original react-google-maps, and this functionality is one of main reasons why I started this project at first place.

FredyC: Ok, was just saying, but it should work nonetheless. You need to kinda try it, that's the purpose of this Next approach to see if it can work without that messy cleanup.

FredyC:
Ok, next batch of changes. I've consolidated that url making function so it's shared and took care of both states when API key and ClientID or none of those are specified.

I've also documented these changes slightly. For some reason the styleguidist needs to have TSX extension for a hook otherwise it doesn't bother with loading it.

I wanted to document the long pending `useGoogleMap` as well, but styleguidist has a problem with the name of file `map-context` and won't load it even if renamed to TSX.

packages/react-google-maps-api/src/LoadScript.tsx
```diff
@@ -179,27 +174,9 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
...
-    }
-
-    if (this.props.libraries && this.props.libraries.length) {
-      params.push(`&libraries=${this.props.libraries.join(",")}`)
-    }
-
     const injectScriptOptions = {
-      id: this.props.id,
-      url: `https://maps.googleapis.com/maps/api/js?${params.join('&')}`
+      id: this.props.id!,
```
JustFly1984: why `!` ?

FredyC: Because of stupid class & defaultProps :) TypeScript is not able to see that default value is assigned so it will never be undefined.

FredyC: For the reference, it was required a prop before, but since it has default, there is no point in requiring it.

JustFly1984: @FredyC 
I have issues with running gatsby-example locally after yarn workspaces merge, can you try to run it, you will see what I mean.

PR #163, opened 2019-06-18
innerText undefined check in slice method

upendra17021989:
# Please explain PR reason.
innerText undefined check in slice method


packages/react-google-maps-api/src/LoadScript.tsx
```diff
@@ -161,7 +161,7 @@ class LoadScript extends React.PureComponent<LoadScriptProps, LoadScriptState> {
     Array.prototype.slice
       .call(document.getElementsByTagName("style"))
       .filter(function filter(style: HTMLStyleElement): boolean {
-        return style.innerText.length > 0 && style.innerText.includes(".gm-")
+        return style.innerText != undefined && style.innerText.length > 0 && style.innerText.includes(".gm-")
```
JustFly1984: please use strict inequality check `!==` instead of `!=`

PR #164, opened 2019-06-18
Use js-fetch for useLoadScript

FredyC:
So after some short use of `useLoadScript` it turned out that obnoxious `google is not defined` is still present. It seems that `onLoad` simply fires too soon before the script is actually processed and attached to `window.google`.

I tried switching to `scriptjs` which does some extra magic and it seems that it fixed the problem. I have tried to investigate what are they doing differently, but given up. Unfortunately, then I noticed it does not support handling load errors.

I found [js-fetch](https://github.com/nghiepit/js-fetch) instead which even has google maps api in readme example, so that should make it like a right tool for the job. It actually checks if `window.google` is present before resolving. As a bonus, the `useLoadScript` returns the `google` object so it's possible to access it without awkward `ts-ignore` when accessing `window`.

### DIFF
--- a/packages/react-google-maps-api/package.json
+++ b/packages/react-google-maps-api/package.json
@@ -83,7 +83,8 @@
   },
   "dependencies": {
     "@react-google-maps/marker-clusterer": "1.2.6",
-    "invariant": "2.2.4"
+    "invariant": "2.2.4",
+    "js-fetch": "^1.3.1"
   },
   "peerDependencies": {
     "react": "^16.6.3",
--- a/packages/react-google-maps-api/src/useLoadScript.tsx
+++ b/packages/react-google-maps-api/src/useLoadScript.tsx
@@ -1,12 +1,12 @@
 import * as React from 'react'
 
-import { injectScript } from './utils/injectscript'
 import { preventGoogleFonts } from './utils/prevent-google-fonts'
 
 import { isBrowser } from './utils/isbrowser'
 import { defaultLoadScriptProps } from './LoadScript'
 import invariant from 'invariant';
 import { makeLoadScriptUrl, LoadScriptUrlOptions } from './utils/make-load-script-url';
+import jsFetch from 'js-fetch'
 
 export interface UseLoadScriptOptions extends LoadScriptUrlOptions {
   id?: string
@@ -50,6 +50,8 @@ export function useLoadScript({
 
   const url = makeLoadScriptUrl({ version, googleMapsApiKey, googleMapsClientId, language, region, libraries })
 
+  const loadedGoogleRef = React.useRef<typeof google>()
+
   React.useEffect(function loadScriptAndModifyLoadedState() {
     if (!isBrowser) {
       return
@@ -67,18 +69,12 @@ export function useLoadScript({
       return
     }
 
-    injectScript({ id, url })
-      .then(setLoadedIfMounted)
-      .catch(function handleInjectError(err) {
-        if (isMounted.current) {
-          setLoadError(err)
-        }
-        console.warn(`
-        There has been an Error with loading Google Maps API script, please check that you provided correct google API key (${googleMapsApiKey || '-'}) or Client ID (${googleMapsClientId || '-'})
-        Otherwise it is a Network issue.
-      `)
-        console.error(err)
-      })
+    jsFetch(url, 'google', { async: true, id }, 5000).then((google: any) => {
+      loadedGoogleRef.current = google
+      setLoadedIfMounted()
+    }).catch((err: Error) => {
+      setLoadError(err)
+    })
   }, [id, url])
 
   const prevLibraries = React.useRef<undefined | string[]>()
@@ -92,5 +88,5 @@ export function useLoadScript({
     prevLibraries.current = libraries
   }, [libraries])
 
-  return { isLoaded, loadError, url }
+  return { isLoaded, loadError, url, google: loadedGoogleRef.current }
 }
--- a/packages/react-google-maps-api/src/utils/js-fetch.d.ts
+++ b/packages/react-google-maps-api/src/utils/js-fetch.d.ts
@@ -1,0 +1,1 @@
+declare module 'js-fetch'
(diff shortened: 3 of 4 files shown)


FredyC:
Huh, what a strange build fail in the gatsby. Is it such a problem to use `eval`? :)

FredyC:
Well, I think I will just gracefully steal the source of that module and include it in here. They don't provide any other than UMD build and apparently Gatsby has a problem with that. Not a worth the hassle.

coulson84:
I was also having issues with useLoadScript not working as expected. my problems were specifically restricted to where useLoadScript was called more than once. 

On the first invocation it creates the script and as all the appropriate handlers for detecting if it loads. However, on every invocation after that it simply checks for the existence of an element with a matching ID and resolves straight away if it is found whether it has loaded or not.

It looks like a bug as it should probably use the readyState attribute of the script to check whether to resolve immediately.

I will submit a PR to that effect if get time.

FredyC:
> It looks like a bug as it should probably use the readyState attribute of the script to check whether to resolve immediately.

Well, the `readyState` is kinda sparsely documented in regards to `script` element. I even found [this answer](https://stackoverflow.com/questions/25617535/script-readystate-is-undefined-in-ie11-ff) that it's no longer supported. I wouldn't be trusting it really.

The js-fetch approach seems more legit, so I would stick to that, just need to figure how to make the build passing. So please hold with your PR for now, we would be conflicting with each other :)

coulson84:
Ah yeah - showing my age there :D Still probably best to avoid adding another dependency just to load a script in. Modifying the existing behaviour of injectScript to fix the issue should be easy enough. One option is below. It's not necessarily ideal. It stores the script status on an attribute. It might be preferable to use an object or map instead of keeping the state on a dom element but this will avoid the extra dependency

```
export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
    if (!isBrowser) {
        return Promise.reject(new Error("document is undefined"))
    }

    return new Promise(function injectScriptCallback(resolve, reject) {
        const scriptEl = document.getElementById(id);

        function addListeners(script) {
            script.addEventListener('load', function onload() {
                script.setAttribute('data-status', 'loaded')
                script.removeEventListener('load', onload)
                resolve(id)
            })

            script.addEventListener('error', function onerror(e) {
                script.setAttribute('data-status', 'error')
                script.removeEventListener('error', onerror)
                reject(e)
            })

        }

        if (scriptEl) {
            const scriptState = scriptEl.getAttribute('data-status')
            if(scriptState === 'loaded') {
                return resolve(id)
            } else if(scriptState === 'loading') {
                return addListeners(script)
            }
            // allow to progress on error state so that it attempts to load the script element again
        }

        const script = document.createElement("script")

        script.type = "text/javascript"
        script.src = url
        script.id = id
        script.async = true
        script.setAttribute('data-status', 'loading')
        addListeners(script);

        document.head.appendChild(script)
    })
        // eslint-disable-next-line @getify/proper-arrows/name
        .catch(err => {
            console.error('injectScript error: ', err)
        })
}
```

FredyC:
@coulson84 I am not too convinced that such an approach is enough. Sure, it would solve the issue with the double use of that hook, but I am fairly convinced it's not enough. There is surely some race condition when the load event does not effectively mark that script has been processed properly. It should make sure to wait for `window.google` to become available.

coulson84:
window.google will always be available when the load event fires on the script element so there should be no need to do a check for window.google first. The only time it would not be present is if the script failed to load but then then error event would trigger and reject the promise.

Give it a try
```
const a = document.createElement('script');
a.src = `https://maps.googleapis.com/maps/api/js?key=${ KEY }&v=weekly&libraries=geometry,drawing,places`;
a.onload = () => alert('window.google is defined = ' + !!window.google);
document.body.append(a);
```
Even if you clear your cache and emulate a slow 3g connection (assuming that any race condition would be due to slow network request) window.google is still defined when the onload event fires because it is created by the script so will always be present at the time the onload event fires.

Furthermore js-fetch doesn't check to see if window.google is defined before resolving with the value it just resolves the returned promise with window.google event if it is undefined. Additionally looks as though it would also suffer from issues if a script failed to load first time and then you tried to load the same script again it would never load it - you can see it pushes items to the urls array but never removes then if the script's error event fires so it would baulk if you tryed to load the same script again.

FredyC:
In the ideal condition, I am sure it works within the `onload` just fine. But I've been running into this issue when running Cypress tests against the app. It is much faster there and many times it ended up with that error and I am 100% certain it wasn't double load or some sort of reuse.

Have you found any sort of official documentation confirming that `onload` is truly final? I can imagine that different browsers can have a slightly different meaning if it's not in W3C specification. Those tests are also executed in headless mode in Elektron based browser. It might be different there as well.

Checking for `window.google` is the safety net in case it's not present at the time of `onload`.

You are right about `js-fetch`. I've looked more carefully and that other block checking with `setInterval` seems useless. It will execute only when requesting the same url again.

coulson84:
Is this PR is now defunct and superseded by https://github.com/JustFly1984/react-google-maps-api/pull/188 or are they distinct issues?

In my experience the onload is handled the same by all browsers. In a previous job I wrote and maintained the JavaScript templates for deploying ads to pages for an adserving company. We had millions of ad impressions everyday and made extensive use of the onload event to serve those ads. I never saw an issue in any browsers (from IE8 onwards) where the onload event fired prior to a script being downloaded, parsed and executed. Our ad metrics would also have indicated a problem in regard to the onload event not working as expected if we were receiving impressions for a certain browser but then no ad loaded or interaction events from it (due to the onload event not working). I guess I am not saying that it will never happen (never say never) but I've never seen it.

Here is a little script that takes a long time to execute when it is loaded on page

```
var script = `
var x = 10000;
var y = {};
while(x--) {
 y = {
   ...y,
   [x]: x,
};
}
`;
var b = new Blob([script], { type: 'text/javascript' });
var u = URL.createObjectURL(b);
var s = document.createElement('script');
s.src = u;
s.addEventListener('load', () => console.log('load') || console.timeEnd('a'));
console.time('a');
document.body.appendChild(s);
document.body.removeChild(s);
URL.revokeObjectURL(u);
```

You can see that the onload console.log doesn't fire until the script finishes running and timer shows a long time (about 14s on my computer). I tested it in Cypress (not in headless mode though) and it showed the same console.log/console.time results as in Chrome pretty much. I don't see any way at all that window.google.maps could be undefined after the onload event fires on the script so I would suspect that the bug is not caused by when the onload event fires but rather some other part of the code.

FredyC:
You are right, this PR is obsolete now, but I think that #187 has solved it in a better way. It makes sense that async loaded script should be notified by the other side when it's done.

PR #180, opened 2019-07-11
Adds ability to pass map instance into the GoogleMaps component

quinnbaetz:
# Please explain PR reason.
Google charges for each instance of a map created: https://cloud.google.com/maps-platform/pricing/sheet/  

Allowing a user to pass in an existing instance adds flexibility to the library.

### DIFF
--- a/packages/react-google-maps-api/src/GoogleMap.tsx
+++ b/packages/react-google-maps-api/src/GoogleMap.tsx
@@ -79,6 +79,7 @@ export interface GoogleMapProps {
   center?: google.maps.LatLng | google.maps.LatLngLiteral;
   clickableIcons?: boolean;
   heading?: number;
+  map?: google.maps.Map;
   mapTypeId?: string;
   streetView?: google.maps.StreetViewPanorama;
   tilt?: number;
@@ -117,7 +118,7 @@ export class GoogleMap extends React.PureComponent<GoogleMapProps, GoogleMapStat
 
   // eslint-disable-next-line @getify/proper-arrows/this, @getify/proper-arrows/name
   getInstance = (): google.maps.Map | null => {
-    return new google.maps.Map(this.mapRef, this.props.options)
+    return this.props.map || new google.maps.Map(this.mapRef, this.props.options)
   }
 
   // eslint-disable-next-line @getify/proper-arrows/this, @getify/proper-arrows/name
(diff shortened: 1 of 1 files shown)


packages/react-google-maps-api/src/GoogleMap.tsx
```diff
@@ -132,6 +133,11 @@ export class GoogleMap extends React.PureComponent<GoogleMapProps, GoogleMapStat
   componentDidMount() {
     const map = this.getInstance()
 
+    if(this.props.map){
+      this.mapRef.appendChild(myFormDomElement)
```
JustFly1984: Your code doesn't work.
It fails on TS check with errors:

```
4:17:19 AM:  FAIL  Failed to compile
4:17:19 AM: [at-loader] ./src/GoogleMap.tsx:137:7
4:17:19 AM:     TS2531: Object is possibly 'null'.
4:17:19 AM: [at-loader] ./src/GoogleMap.tsx:137:31
4:17:19 AM:     TS2304: Cannot find name 'myFormDomElement'.
4:17:19 AM: [at-loader] ./src/GoogleMap.tsx:138:7
4:17:19 AM:     TS2531: Object is possibly 'null'.
4:17:19 AM: [at-loader] ./src/GoogleMap.tsx:138:22
4:17:19 AM:     TS2345: Argument of type 'MapOptions | undefined' is not assignable to parameter of type 'MapOptions'.
4:17:19 AM:   Type 'undefined' is not assignable to type 'MapOptions'.
4:17:20 AM: error Command failed with exit code 1.
```

quinnbaetz: Yep, my bad, just copied over from a shim I had.  Do you think in theory this is a useful thing that should be in the library?

JustFly1984: @quinnbaetz
I think your code does not execute what you want. The whole google-maps-api thing is pretty buggy, and there is no open source. I'm not sure that you can reuse the same map over and over again for several maps on the same page. Also it could cause memory leaks on mount/unmount stages. Please read source code, and test first. You can run `yarn tc` to check if project builds. Also if you use VScode for editing - it has typescript support out of the box.

quinnbaetz: 👍 We are using the same map over and over again on the same page, but only one will be shown at a time.  I will revisit this when I have some more time.  Closing PR for now.

PR #183, opened 2019-07-16
Fixed #182 - onOverlayComplete not called 

### LINKED ISSUE #182: DrawingManager onOverlayComplete not firing when setting drawingMode null
Using the native Google Maps API, while drawing a polygon, setting `drawingManager.setDrawingMode(null)` fires the 'overlaycomplete' event.

Using react-google-maps-api, this does not appear to be the case.

Given boilerplate like this:

```js
const [cancelFlag, setCancelFlag] = useState(false)
const [drawingMode, setDrawingMode] = useState(null)

const handleOverlayComplete = (e) => {
  if (cancelFlag) {
    e.overlay.setMap(null) // deletes the shape
  } else {
    // do things with the shape ...
  }

const onClickOutsideMap = () => {
  setCancelFlag(true)
  setDrawingMode(null)
}

const onPressEscapeKey = () => onClickOutsideMap()
...
<DrawingManager
  drawingMode={drawingMode}
  onOverlayComplete={handleOverlayComplete}
  ...
```

Two handlers, `onClickOutsideMap` and `onPressEscapeKey`, do `setCancelFlag(true)`, then `setDrawingMode(null)`. 

The latter, I would expect to trigger the onOverlayComplete event, which **IS** the case when using the native Google Maps API. However, `setDrawingMode(null)` does **NOT** trigger the 'overlaycomplete' event using react-google-maps-api.

The purpose of the above is to cancel a polygon mid-draw.

**With react-google-maps-api, `setDrawingMode(null)` mid-draw causes the shape to "auto-complete" -- the last point connects to the first, completing the shape -- but the 'overlaycomplete' event does not fire.**

uriklar:
current  - Events are unregistered, then updates are applied, then evens are re-registered
This causes a bug that if an update needs to cause an event call, that event isn't registered at the time of the update. (as seen in #182)

Switching the order solves this issue

### DIFF
--- a/packages/react-google-maps-api/src/utils/helper.ts
+++ b/packages/react-google-maps-api/src/utils/helper.ts
@@ -75,6 +75,7 @@ export function applyUpdatersToPropsAndRegisterEvents({
   nextProps: any;
   instance: any;
 }) {
+  const registeredEvents = registerEvents(nextProps, instance, eventMap)
   applyUpdaterToNextProps(updaterMap, prevProps, nextProps, instance)
-  return registerEvents(nextProps, instance, eventMap)
+  return registeredEvents
 }


mjcampagna:
@uriklar Is this in the 1.4.2 release? If so, `overlaycomplete` still seems not to be fired from the onOverlayComplete prop when changing the drawing mode.

Issue #184, opened 2019-07-17
`onload` doesn't confirm that `google` is initialized!


ReiiYuki:
Hi

https://github.com/JustFly1984/react-google-maps-api/blob/5dd1be2cf435f5ab5f711ff63e86ff6eabd0ee43/packages/react-google-maps-api/src/utils/injectscript.ts#L24

`onload`  doesn't confirm that `google` is initialized!

```html
  <script async defer src="https://maps.googleapis.com/maps/api/js?key=YOUR_API_KEY&callback=initMap"
  type="text/javascript"></script>
```

I think you should resolve loaded status by trigger some callback function like initMap.

JustFly1984:
@ReiiYuki  Can you make a PR please?

ReiiYuki:
@JustFly1984 OK! now it's on #187

baldmountain:
I just ran into this. Sometimes google is ready, but most times not.

ReiiYuki:
@JustFly1984 From the latest MR I found that it doesn't cover every case of loaded script.

The case I found is I use `useLoadScript` in multiple components.

The logic of `injectScript` which is used in `useLoadScript` is the following:
- if existing script tag resolve it as loaded!
- otherwise create new script tag then make it resolve when callback (`initMap`) is called.

This mean if we call useLoadScript with the same id multiple times, if `initMap` doesn't call yet, it will resolve as loaded because it found existing script tag.

ReiiYuki:
I suggest that we should check if script is already loaded by create new window attribute as save the value of loaded id in that attribute.

JustFly1984:
@ReiiYuki can you make a PR?

coulson84:
Have added a PR here https://github.com/JustFly1984/react-google-maps-api/pull/194 that should fix this

PR #187, opened 2019-07-17
Use callback instead of onload to resolve injectScript as loaded!

ReiiYuki:
# Please explain PR reason.

According #184 

`onload` of `script` DOM will be call after the script is loaded.

However it doesn't mean that `google` is ready to be used in your app.

According Google Map Javascript document, it recommends to be use `callback` function to continue the process after `google` is ready.

In this PR, I created `initMap` function as global function for make it resolve script after `google` is ready in your app.

### DIFF
--- a/packages/react-google-maps-api/src/utils/injectscript.ts
+++ b/packages/react-google-maps-api/src/utils/injectscript.ts
@@ -9,8 +9,6 @@ interface InjectScriptArg {
   id: string;
 }
 
-const windowWithGoogleMap: WindowWithGoogleMap = window
-
 export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
   if (!isBrowser) {
     return Promise.reject(new Error("document is undefined"))
@@ -27,6 +25,11 @@ export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
       else {
         existingScript.remove()
       }
+    }
+    const windowWithGoogleMap: WindowWithGoogleMap = window
+
+    if (document.getElementById(id)) {
+      return resolve(id)
     }
 
     const script = document.createElement("script")
--- a/packages/react-google-maps-api/src/utils/injectscript.ts
+++ b/packages/react-google-maps-api/src/utils/injectscript.ts
@@ -1,9 +1,15 @@
 import { isBrowser } from "./isbrowser"
 
+interface WindowWithGoogleMap extends Window {
+  initMap?: () => void
+}
+
 interface InjectScriptArg {
   url: string;
   id: string;
 }
+
+const windowWithGoogleMap: WindowWithGoogleMap = window
 
 export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
   if (!isBrowser) {
@@ -29,10 +35,11 @@ export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
     script.src = url
     script.id = id
     script.async = true
-    script.onload = function onload() {
+    script.onerror = reject
+
+    windowWithGoogleMap.initMap = function onload() {
       resolve(id)
     }
-    script.onerror = reject
 
     document.head.appendChild(script)
   })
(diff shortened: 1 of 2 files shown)


FredyC:
Sound interesting. Where in docs is the `callback` mentioned? I don't recall seeing anything like that in there. If it really works, then it LGTM.

JustFly1984:
@ReiiYuki can you please provide the link to the docs? @uriklar @FredyC can you please test?

ReiiYuki:
This one,

https://developers.google.com/maps/documentation/javascript/tutorial

`callback` is mentioned in section called `Synchronously loading the API`

FredyC:
I think it makes total sense to do this. We have an `async` mode enabled there, so it's logical to have some sort of feedback it's done.

packages/react-google-maps-api/src/utils/injectscript.ts
```diff
@@ -22,17 +26,23 @@ export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
         existingScript.remove()
       }
     }
+    const windowWithGoogleMap: WindowWithGoogleMap = window
+
+    if (document.getElementById(id)) {
```
slorber: I'm afraid this will reintroduce a bad behavior that was fixed here: https://github.com/JustFly1984/react-google-maps-api/pull/189/files

slorber:
Makes sense  for me too. Google maps seems to load additional resources after first script is loaded so it's probably better to wait for google maps to notify us when ready

Take care of not reverting behavior of this recent merged PR: https://github.com/JustFly1984/react-google-maps-api/pull/189/files

packages/react-google-maps-api/src/utils/injectscript.ts
FredyC: Well, if we enforce a single static ID, there is no problem.

slorber: There could, because if the url change and we won't create the new script.

But looking at the whole picture: https://github.com/JustFly1984/react-google-maps-api/blob/20194bba212bc641d981d2be9fb679f46f05bf7a/packages/react-google-maps-api/src/utils/injectscript.ts

This line is actually useless, because when going to this line, the script will weither not exist or has been removed, so `document.getElementById(id)` will never return a script

baldmountain:
Hi, I just ripped out tomchentw's version of react-google-maps and replaced it with this library. I kind of need this fix since the library is failing just about every time I load a map with the error that window.google doesn't exist.

JustFly1984:
Have you tested this, can I merge and publish? I just have no time to test, but I can merge and publish 1.5.0 with this PR

FredyC:
No time to test either, but considering it's documented, I think it's safe to merge.

JustFly1984:
@baldmountain @FredyC @slorber @ReiiYuki 
I have published 1.5.0 and updated gatsby examples. please test and your feedback is welcome)

ReiiYuki:
Thanks!

baldmountain:
This didn't work for me at all. I'm using useLoadScript and it's failing saying that a hook is being called outside a component function.

I can't fuss with this anymore. I'll just use tomchentw`s version since it's working for me.

JustFly1984:
@baldmountain Where do you call the hook? can you provide an example at codesandbox?

FredyC:
@baldmountain Note that you can use `LoadScriptNext` if you have trouble understanding hooks.

baldmountain:
@JustFly1984   Literally the first line inside the component method. (The stateless component that contains the GoogleMap component.)

I should mention that the component is not visible when the page loads. When a user clicks "View Map" the Map component becomes visible and renders.

JustFly1984:
@baldmountain 
which component method?

stateless functional components has no methods :-)

FredyC:
@baldmountain Better to show a reproduction instead of trying to describe it ;)

baldmountain:
This is quick and dirty but should be roughly what my component looks like: https://codesandbox.io/s/wonderful-sun-2jqp2

You'll need to set your own google maps key in `YOUR_API_KEY`

This actually get a CORS error, but the structure should be correct.

Issue #186, opened 2019-07-17
Dynamically changing language or libraries


slorber:
Hi,

On 1.4.2

It is complicated to change at runtime the language or libraries supported. Clearly it's not the fault of this library, google does not make it easy. 

One can't simply pass a language prop to the map or places autocomplete to customize to the needed language/libraries, we have to reload the global google script with the correct language/libraries.

Even in this page, the demo which allow to change languages do that (reloading the page with a new script with language in querystring): https://developers.google.com/maps/documentation/javascript/localization

The workaround to solve this issue is:
- remove old google script
- add new one with a new url

Demo: http://jsfiddle.net/2AKqM/

--------------------------

I think it would be cool if this lib made all this process easier.

As it's not possible to use multiple languages on the same page, I think:
- The lib should recommend loading the script at most once at the same time
- The lib should ensure at most one google script is in <head>, and eventually do the cleanup

Currently, if the language change, the url will change, and we will try to insert a new script but this will be [prevented](https://github.com/JustFly1984/react-google-maps-api/blob/5dd1be2cf435f5ab5f711ff63e86ff6eabd0ee43/packages/react-google-maps-api/src/utils/injectscript.ts#L14), unless we change the id of the script.

But if we change the id of the script (for example, by putting language and libraries in it), then the older script tags remains in the dom and are not cleaned. The more we use the app and the more scripts get loaded over time and no cleanup is done.

I don't have a proper solution to this problem yet but would like to have your insight on this topic

-----

Related:
- https://github.com/JustFly1984/react-google-maps-api/issues/186
- https://github.com/JustFly1984/react-google-maps-api/pull/188
- https://github.com/JustFly1984/react-google-maps-api/pull/189

slorber:
Also, if we call again injectScript (due to a changing url/id) the state "loaded" remains true. I think in cases we are loading a new script, we want to reset the state so that we can show a spinner on something on our apps

JustFly1984:
The language change does reloads the script, but libraries change does not.

slorber:
@JustFly1984 not sure what you mean.

The libraries are included in the url:
https://github.com/JustFly1984/react-google-maps-api/blob/master/packages/react-google-maps-api/src/utils/make-load-script-url.ts#L9

Which means if the libraries chance, the url change, and the script reloads.

Is there another way to add a maps library dynamically without reloading the script?

JustFly1984:
[...]

PR #188, opened 2019-07-18
fix useLoadScript not resetting its state when id/url is changed

slorber:
# Please explain PR reason.

I think if a new script gets inserted, the "loaded: boolean" state should go back to false so that we can wait in consumer code for new script to load

cc @JustFly1984 @FredyC 

Related:
- https://github.com/JustFly1984/react-google-maps-api/issues/186
- https://github.com/JustFly1984/react-google-maps-api/pull/188
- https://github.com/JustFly1984/react-google-maps-api/pull/189

### DIFF
--- a/packages/react-google-maps-api/src/useLoadScript.tsx
+++ b/packages/react-google-maps-api/src/useLoadScript.tsx
@@ -8,6 +8,7 @@ import { preventGoogleFonts } from './utils/prevent-google-fonts'
 import { makeLoadScriptUrl, LoadScriptUrlOptions } from './utils/make-load-script-url'
 
 import { defaultLoadScriptProps } from './LoadScript'
+import { useCallback } from 'react'
 
 export interface UseLoadScriptOptions extends LoadScriptUrlOptions {
   id?: string;
@@ -15,6 +16,17 @@ export interface UseLoadScriptOptions extends LoadScriptUrlOptions {
 }
 
 let previouslyLoadedUrl: string
+
+const useIsMounted = () => {
+  const isMountedRef = React.useRef(false)
+  React.useEffect(function trackMountedState() {
+    isMountedRef.current = true
+    return () => {
+      isMountedRef.current = false
+    }
+  }, [])
+  return useCallback(() => isMountedRef.current,[]);
+}
 
 export function useLoadScript({
   id = defaultLoadScriptProps.id,
@@ -26,16 +38,9 @@ export function useLoadScript({
   libraries,
   preventGoogleFontsLoading,
 }: UseLoadScriptOptions) {
-  const isMounted = React.useRef(false)
+  const isMounted = useIsMounted();
   const [isLoaded, setLoaded] = React.useState(false)
   const [loadError, setLoadError] = React.useState<Error | undefined>(undefined)
-
-  React.useEffect(function trackMountedState() {
-    isMounted.current = true
-    return () => {
-      isMounted.current = false
-    }
-  }, [])
 
   React.useEffect(function applyPreventGoogleFonts() {
     if (isBrowser && preventGoogleFontsLoading) {
@@ -60,8 +65,15 @@ export function useLoadScript({
       return
     }
 
+    function resetStateIfMounted() {
+      if (isMounted()) {
+        setLoaded(false)
+        setLoadError(undefined)
+      }
+    }
+
     function setLoadedIfMounted() {
-      if (isMounted.current) {
+      if (isMounted()) {
         setLoaded(true)
         previouslyLoadedUrl = url
       }
@@ -72,10 +84,12 @@ export function useLoadScript({
       return
     }
 
+    resetStateIfMounted();
+
     injectScript({ id, url })
       .then(setLoadedIfMounted)
       .catch(function handleInjectError(err) {
-        if (isMounted.current) {
+        if (isMounted()) {
           setLoadError(err)
         }
         console.warn(`
--- a/packages/react-google-maps-api/src/useLoadScript.tsx
+++ b/packages/react-google-maps-api/src/useLoadScript.tsx
@@ -9,13 +9,24 @@ import { makeLoadScriptUrl, LoadScriptUrlOptions } from './utils/make-load-scrip
 
 import { defaultLoadScriptProps } from './LoadScript'
 import useIsMounted from './utils/useIsMounted'
+import { createUseAtMostOnce } from './utils/useAtMostOnce'
+import usePrevious from './utils/usePrevious'
 
 export interface UseLoadScriptOptions extends LoadScriptUrlOptions {
-  id?: string;
-  preventGoogleFontsLoading?: boolean;
+  id?: string
+  preventGoogleFontsLoading?: boolean
 }
 
 let previouslyLoadedUrl: string
+
+const useLoadScriptAtMostOnce = createUseAtMostOnce("useLoadScript should only be used at most once. The google script is loaded globally, and one hook's load could override another hook's load")
+
+const useIdNeverChange = (id: string) => {
+  const previousId = usePrevious(id) || id;
+  if (id !== previousId) {
+    throw new Error('useLoadScript does not support changing dynamically the script id')
+  }
+}
 
 export function useLoadScript({
   id = defaultLoadScriptProps.id,
@@ -25,9 +36,12 @@ export function useLoadScript({
   language,
   region,
   libraries,
-  preventGoogleFontsLoading,
+  preventGoogleFontsLoading
 }: UseLoadScriptOptions) {
-  const isMounted = useIsMounted();
+  useLoadScriptAtMostOnce()
+  useIdNeverChange(id);
+
+  const isMounted = useIsMounted()
   const [isLoaded, setLoaded] = React.useState(false)
   const [loadError, setLoadError] = React.useState<Error | undefined>(undefined)
 
@@ -55,14 +69,14 @@ export function useLoadScript({
     }
 
     function resetStateIfMounted() {
-      if (isMounted()) {
+      if (isMounted.current) {
         setLoaded(false)
         setLoadError(undefined)
       }
     }
 
     function setLoadedIfMounted() {
-      if (isMounted()) {
+      if (isMounted.current) {
         setLoaded(true)
         previouslyLoadedUrl = url
       }
@@ -78,7 +92,7 @@ export function useLoadScript({
     injectScript({ id, url })
       .then(setLoadedIfMounted)
       .catch(function handleInjectError(err) {
-        if (isMounted()) {
+        if (isMounted.current) {
           setLoadError(err)
         }
         console.warn(`
--- a/packages/react-google-maps-api/src/useLoadScript.tsx
+++ b/packages/react-google-maps-api/src/useLoadScript.tsx
@@ -8,7 +8,7 @@ import { preventGoogleFonts } from './utils/prevent-google-fonts'
 import { makeLoadScriptUrl, LoadScriptUrlOptions } from './utils/make-load-script-url'
 
 import { defaultLoadScriptProps } from './LoadScript'
-import { useCallback } from 'react'
+import useIsMounted from './utils/useIsMounted'
 
 export interface UseLoadScriptOptions extends LoadScriptUrlOptions {
   id?: string;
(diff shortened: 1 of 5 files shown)


packages/react-google-maps-api/src/useLoadScript.tsx
```diff
@@ -16,6 +17,17 @@ export interface UseLoadScriptOptions extends LoadScriptUrlOptions {
...
 
+const useIsMounted = () => {
+  const isMountedRef = React.useRef(false)
+  React.useEffect(function trackMountedState() {
+    isMountedRef.current = true
+    return () => {
+      isMountedRef.current = false
+    }
+  }, [])
+  return useCallback(() => isMountedRef.current,[]);
```
FredyC: Why callback? It does need referential integrity. I think it's better to just return a whole ref.

FredyC: And I would probably move this whole hook into a separate file. I can imagine it will find its use in other places eventually.

packages/react-google-maps-api/src/useLoadScript.tsx
```diff
@@ -72,10 +84,12 @@ export function useLoadScript({
       return
     }
 
+    resetStateIfMounted();
```
FredyC: Please separate this logic into its own useEffect. No need to reset when id changes. Also, it's better for clarity of intention.

slorber: Actually, if id change and url does not change, injectScript will always insert a new script (which imho is not so good because the older script remains, we should provide a strategy to update the id of the existing script instead?)

So basically, I really want to call this resetState whenever injectScript() is called

packages/react-google-maps-api/src/useLoadScript.tsx
slorber: About ref, I personnally prefer returning a isMounted() function, which seems to be simpler to manipulate as it's just a js concept, not a react one). Don't remember who said that but "use the least powerful abstraction that gets the job done". I guess you meant referential transparency, because integrity seems related to RDBMS for me.

Will extract that in a separate file

packages/react-google-maps-api/src/useLoadScript.tsx
FredyC: Frankly, I never understood why we even allow customizing the ID considering, that all will always be set to `window.google`. There is no point in having multiple configurations as it will always overwrite the previous one.

packages/react-google-maps-api/src/utils/useIsMounted.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+const useIsMounted = () => {
+  const isMountedRef = React.useRef(false)
+  React.useEffect(function trackMountedState() {
+    isMountedRef.current = true
+    return () => {
+      isMountedRef.current = false
+    }
+  }, [])
+  return () => isMountedRef.current
```
FredyC: Please, let's drop a function completely. There is no benefit of hiding it in there and on contrary, you need to keep in deps of callbacks & effects. When you return the whole ref, it will always have the latest value

packages/react-google-maps-api/src/useLoadScript.tsx
JustFly1984: @FredyC 
id is required to find `<script />` element in DOM to track if script mounted or not

FredyC: Yea sure, but it can be static, right? There is no reason to keep it configurable and a source of confusion. Or at least consider it only initially and don't worry about it when it changes.

slorber: Agree, I don't see an usecase for having that id change over time. Used this only as a workaround to trigger script updates when injectScript did not reinject on url changes (fixed by my merged PR).

We should rather error if user tries to modify id dynamically imho, because allowing changing id just makes things a bit more complicated for nothing

slorber: Also, if we could find a way to ensure that at most one hook is mounted, that would ensure more guarantees as well. The user should rather use the hook upper in his tree and only at most once, instead of colocating the load into multiple pages.

For nested/local spinners, one can propagate isLoaded through context

packages/react-google-maps-api/src/utils/useIsMounted.tsx
slorber: ok, but that was actually the point of useCallback in the first place to return a constant value over time.

But if you prefer a ref i really dont care much

packages/react-google-maps-api/src/useLoadScript.tsx
FredyC: Mostly agree, just loader shouldn't be a single one on top because in bigger apps you don't want to be loading map resources before you know they are really needed. Especially on mobile connections.

It could be solved if we add `enabled` to options and then it's possible to drive that through an app state and leave it disabled until really needed.

packages/react-google-maps-api/src/utils/useAtMostOnce.ts
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react'
+import { useEffect } from 'react'
+
+export const createUseAtMostOnce = (
```
FredyC: I doubt this is gonna work. The refs are like instance properties in classes. They are not shared. You would need a global variable which is bad. Let's not force this check upon the user.

packages/react-google-maps-api/src/useLoadScript.tsx
```diff
@@ -9,14 +9,25 @@ import { makeLoadScriptUrl, LoadScriptUrlOptions } from './utils/make-load-scrip
...
 export interface UseLoadScriptOptions extends LoadScriptUrlOptions {
-  id?: string;
-  preventGoogleFontsLoading?: boolean;
+  id?: string
+  preventGoogleFontsLoading?: boolean
 }
 
 let previouslyLoadedUrl: string
 
+const useLoadScriptAtMostOnce = createUseAtMostOnce("useLoadScript should only be used at most once. The google script is loaded globally, and one hook's load could override another hook's load")
```
FredyC: I certainly don't like this. We shouldn't be forcing that upon the user. There is no technical reason why you cannot use it more times. If anything, it can show warning, but it certainly cannot throw an Error.

packages/react-google-maps-api/src/utils/useIsMounted.tsx
FredyC: Point is that even with useCallback you need to have in deps just in case it's recreated for some reason. Refs are different, they have a stable reference by default.

packages/react-google-maps-api/src/useLoadScript.tsx
slorber: if you look, it's actually creating a global variable, it's not using useRef() but creating that global varible hook-less.

But ok if you don't like I'll remove.

So you are ok to let user do something wrong (ie using twice this hook in the same page), and not warning him? (actually GoogleMaps will likely warn him anyway, unless we have a cleanup system to `delete google.maps` on hook unmount)

packages/react-google-maps-api/src/useLoadScript.tsx
```diff
@@ -8,14 +8,26 @@ import { preventGoogleFonts } from './utils/prevent-google-fonts'
...
+
+const useIdNeverChange = (id: string) => {
+  const previousId = usePrevious(id) || id;
```
FredyC: I think we should actually remove that ID from options completely. What's the point of allowing to customize it?

packages/react-google-maps-api/src/useLoadScript.tsx
slorber: on my Gatsby SPA I just put the hook at the page level, not main app root comp. The load will not happen until the user navigates to a page using maps.

Also, I handle cleanup (mostly to avoid the Gmaps error due to loading twice)

```
const useGoogleMapsAPILoader = () => {
  const language = useGoogleMapsLanguage();
  const scriptId = `GoogleMaps_lang=${language}`; // Should not be needed anymore now that injectScript is fixed
  useEffect(
    () => {
      return () => {
        const script = document.getElementById(scriptId);
        if (script) {
          script.remove();
          // @ts-ignore
          if (window.google && window.google.maps) {
            // @ts-ignore
            delete window.google.maps;
          }
        }
      };
    },
    [language],
  );

  return useLoadScript({
    id: scriptId,
    language,
    libraries: AppMapLibraries,
    googleMapsApiKey: AppConfiguration.GoogleMapsApiKey,
  });
}; 
```

packages/react-google-maps-api/src/useLoadScript.tsx
FredyC: I think it's quite a bad pattern. Hooks are generally for reusability purposes, people know it. It shouldn't be hook if it's not intended to be reused.

Cleanup is still in the "old" LoadScript and most likely was a reason for weird behavior and that's why I made this hook. There is no need. Google will replace `window.google` no matter if anything else is there.

packages/react-google-maps-api/src/useLoadScript.tsx
slorber: it works fine, the only thing that annoys me with that system is that if maps has been loaded on page1, and user goes to page 2, then the map will always reload even if the url is the same. So a solution to enable/disable at the root of my app would be useful ;)

packages/react-google-maps-api/src/useLoadScript.tsx
slorber: I have no id :)

I think we are not yet clear on what's the best solution to load this lib correctly. Shouldn't we pursue discussions in issues?

packages/react-google-maps-api/src/useLoadScript.tsx
slorber: We could say exactly the same about components, yet there are components that are not meant to be mounted more than once per app and will throw if you try. For example react-navigation's appContainer.

Google will replace window.google, but with an error everytime, which will popup in my Sentry, which I don't like ;)

JustFly1984: well, there is an easy way to get rid of an error) 

remove `version` prop from `window.google` object

Screenshot 2019-07-19 14 29 11

slorber: Thanks, that looks like an useful tip to know.

What about if we:
- use a constant id
- when script is going to be removed/recreated due to url change, delete previous google.version to automatically avoid the error? That would be safer than deleting google.maps

JustFly1984: I have read the code of google maps, and even it is minified, the error string is searchable, and  error is shown only if `window.google.maps.version` is not undefined

FredyC:
Ok, so after a lengthy discussion at #194 I think this is a good PR. I've also realized that as much as restrictive it looks, it doesn't break anything. As long as there is a single `useLoadScript` at the time, it won't do any harm. That fits my use case just fine I don't even need to extra loading to some top component.

However, about restricting change to `id`, I think we shouldn't do that here as it might be breaking change as @coulson84 pointed out in https://github.com/JustFly1984/react-google-maps-api/pull/194#issuecomment-515934419. Although we could turn it into a _deprecation warning_ here instead of throwing an error for now.

JustFly1984:
can I merge this branch?

FredyC:
Not yet please, see the second portion of my comment.

slorber:
Hi,

Sorry but I don't have much time to work on these problems.

I'm thinking to extract a lib, decoupled from react, to handle those google maps loading problems, that could help the whole js community instead of just react users imho.

The idea would be something like this:
- have a function to inject script take a google maps url
- first call injects the script, and put "count=1" in somewhere
- return a cleanup callback, that will just do count--
- second call with a different url throws because unsafe
- second call with same url just count++
- if count=0 and we inject script again, if same url, just do count++, else delete old script, delete google.maps.version to avoid the console error, and then reinsert the new script with count=1

FredyC:
@slorber Idea is nice, but since you don't have time to fix the PR, it's not very helpful :) I guess I will make a new PR based on your branch a figure it out.

slorber:
sure you can pick my pr and enhance it and I can help review it.

I've a google maps setup a bit hacky but it works, so it's not currently a high priority for me to work on this for my current work and have a lot of other things to handle :s

JustFly1984:
considering this issue closed due to #199

PR #194, opened 2019-07-24
in injectScript don't immediately resolve if matching script is in the dom

coulson84:
When injectScript runs it checks for the presence of a script element with matching ID in the DOM. If it finds it, it immediately resolves, regardless of whether the script has finished loading or not.

This changes add a 'state' attribute to the script that can be used to check if the google maps is ready or not and then use that information to decide whether to resolve immediately.

There are a couple of possible issues. I would probably prefer using `script.addEventListener('load', callback)` instead of `script.onload` and then using the closure to maintain a reference back to the original call to injectScript's onload callback. That's just what I prefer but attempting with changing the source as little as possible to implement the feature.

I have seen in other threads the mention of removing the id from the call but that it was needed to to find the script via document.getElementById. Why not use the below to find scripts according to just the url? I am not sure if this has been considered but found to be not suitable or not?

```
document.querySelector(`script[src="${ url }"]`)
```

### DIFF
--- a/packages/react-google-maps-api/src/__tests__/utils/injectscript.test.ts
+++ b/packages/react-google-maps-api/src/__tests__/utils/injectscript.test.ts
@@ -4,21 +4,7 @@ interface WindowWithGoogleMap extends Window {
   initMap?: () => void;
 }
 
-// interface ScriptWithOnErrorHandler extends Element {
-//     onerror?: (err: Error) => void;
-// }
-
 describe('utils/injectScript', () => {
-  // let urls: string[] = [];
-
-  // function createScriptUrl(script = '(function(){ var a = 1; })()') {
-  //     const b = new Blob([script], { type: 'text/javascript' });
-  //     const url = URL.createObjectURL(b);
-  //     urls.push(url)
-  //     return url;
-  // }
-
-  // afterAll(() => urls.forEach(url => URL.revokeObjectURL(url)));
 
   describe('injectScript', () => {
     it('adds a script to the page with id and src attributes that correspond to the id and url properties', () => {
@@ -59,28 +45,6 @@ describe('utils/injectScript', () => {
       })
         .then(value => expect(value).toBeTruthy())
     })
-
-
-    // it('returns a promise which rejects if the script does not load', () => {
-    //     const id = 'injectScript-test3'
-    //     const url = `http://localhost/script3.js`
-
-    //     setTimeout(() => {
-    //         const script: ScriptWithOnErrorHandler | null = document.querySelector(id);
-    //         console.log(script && script.onerror);
-    //         console.log('onerror');
-    //         if(script && script.onerror) {
-    //             script.onerror(new Error());
-    //         }
-    //     }, 0)
-
-    //     return injectScript({
-    //         id,
-    //         url,
-    //     })
-    //         .then(() => expect('This then should not run').toBe(''))
-    //         .catch(value => expect(value).toBeInstanceOf(Error))
-    // })
 
     describe('duplicate calls with matching id and url properties', () => {
       it('do not add another script element to the page', () => {
--- a/packages/react-google-maps-api/src/__tests__/utils/injectscript.test.ts
+++ b/packages/react-google-maps-api/src/__tests__/utils/injectscript.test.ts
@@ -1,0 +1,143 @@
+import { injectScript } from '../../utils/injectscript'
+
+interface WindowWithGoogleMap extends Window {
+  initMap?: () => void;
+}
+
+// interface ScriptWithOnErrorHandler extends Element {
+//     onerror?: (err: Error) => void;
+// }
+
+describe('utils/injectScript', () => {
+  // let urls: string[] = [];
+
+  // function createScriptUrl(script = '(function(){ var a = 1; })()') {
+  //     const b = new Blob([script], { type: 'text/javascript' });
+  //     const url = URL.createObjectURL(b);
+  //     urls.push(url)
+  //     return url;
+  // }
+
+  // afterAll(() => urls.forEach(url => URL.revokeObjectURL(url)));
+
+  describe('injectScript', () => {
+    it('adds a script to the page with id and src attributes that correspond to the id and url properties', () => {
+      const url = `http://localhost/script1.js`
+      const value = injectScript({
+        id: 'injectScript-test1',
+        url,
+      })
+
+
+      setTimeout(() => {
+        const win: WindowWithGoogleMap = window
+        if(win.initMap) {
+          win.initMap()
+        }
+      }, 0)
+
+      const element = document.querySelector('#injectScript-test1')
+      expect(element).toBeTruthy()
+      expect(element).toHaveProperty('src', url)
+      return value
+    })
+
+    it('returns a promise which resolves when the script loads', () => {
+      const id = 'injectScript-test2'
+      const url = `http://localhost/script2.js`
+
+      setTimeout(() => {
+        const win: WindowWithGoogleMap = window
+        if(win.initMap) {
+          win.initMap()
+        }
+      }, 0)
+
+      return injectScript({
+        id,
+        url,
+      })
+        .then(value => expect(value).toBeTruthy())
+    })
+
+
+    // it('returns a promise which rejects if the script does not load', () => {
+    //     const id = 'injectScript-test3'
+    //     const url = `http://localhost/script3.js`
+
+    //     setTimeout(() => {
+    //         const script: ScriptWithOnErrorHandler | null = document.querySelector(id);
+    //         console.log(script && script.onerror);
+    //         console.log('onerror');
+    //         if(script && script.onerror) {
+    //             script.onerror(new Error());
+    //         }
+    //     }, 0)
+
+    //     return injectScript({
+    //         id,
+    //         url,
+    //     })
+    //         .then(() => expect('This then should not run').toBe(''))
+    //         .catch(value => expect(value).toBeInstanceOf(Error))
+    // })
+
+    describe('duplicate calls with matching id and url properties', () => {
+      it('do not add another script element to the page', () => {
+        const id = 'injectScript-test4'
+        const url = 'http://localhost/script4.js'
+
+        injectScript({
+          id,
+          url,
+        })
+
+        const value = injectScript({
+          id,
+          url,
+        })
+
+
+        setTimeout(() => {
+          const win: WindowWithGoogleMap = window
+          if(win.initMap) {
+            win.initMap()
+          }
+        }, 0)
+
+        expect(document.querySelectorAll(`script[src="${url}"]`)).toHaveProperty('length', 1)
+        return value
+      })
+
+      it('do not resolve in successive calls unless the script has actually loaded from the first call', () => {
+        const id = 'injectScript-test5'
+        const url = 'http://localhost/script5.js'
+        let initMapCalled = false
+
+        const value1 = injectScript({
+          id,
+          url,
+        })
+
+        const value2 = injectScript({
+          id,
+          url
+        })
+
+        setTimeout(() => {
+          const win: WindowWithGoogleMap = window
+          if(win.initMap) {
+            initMapCalled = true
+            win.initMap()
+          }
+        }, 50)
+
+        return value2.then(() => {
+          expect(initMapCalled).toBe(true)
+          return value1
+        })
+      })
+
+    })
+  })
+})
--- a/packages/react-google-maps-api/src/utils/injectscript.ts
+++ b/packages/react-google-maps-api/src/utils/injectscript.ts
@@ -1,7 +1,7 @@
 import { isBrowser } from "./isbrowser"
 
 interface WindowWithGoogleMap extends Window {
-  initMap?: () => void
+  initMap?: () => void;
 }
 
 interface InjectScriptArg {
@@ -16,17 +16,26 @@ export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
 
   return new Promise(function injectScriptCallback(resolve, reject) {
     const existingScript = document.getElementById(id) as HTMLScriptElement | undefined
+    const windowWithGoogleMap: WindowWithGoogleMap = window
     if (existingScript) {
       // Same script id/url: keep same script
       if (existingScript.src === url) {
-        return resolve(id)
+        if(existingScript.getAttribute('state') === 'ready') {
+          return resolve(id)
+        }
+        else {
+          const onGoogleMapsReadyClosure = windowWithGoogleMap.initMap
+          existingScript.onload = function() {
+            onGoogleMapsReadyClosure()
+            resolve(id)
+          }
+        }
       }
       // Same script id but url changed: recreate the script
       else {
         existingScript.remove()
       }
     }
-    const windowWithGoogleMap: WindowWithGoogleMap = window
 
     if (document.getElementById(id)) {
       return resolve(id)
@@ -39,8 +48,10 @@ export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
     script.id = id
     script.async = true
     script.onerror = reject
+    script.setAttribute('state', 'loading')
 
     windowWithGoogleMap.initMap = function onload() {
+      script.setAttribute('state', 'ready')
       resolve(id)
     }
 
--- a/packages/react-google-maps-api/src/utils/injectscript.ts
+++ b/packages/react-google-maps-api/src/utils/injectscript.ts
@@ -25,7 +25,7 @@ export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
         }
         else {
           const onGoogleMapsReadyClosure = windowWithGoogleMap.initMap
-          existingScript.onload = function() {
+          windowWithGoogleMap.initMap = function() {
             if(onGoogleMapsReadyClosure) {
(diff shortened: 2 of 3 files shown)


JustFly1984:
Url is variable, for example if we change language or libraries, it will change

coulson84:
Ok, I must've misinterpreted the requirements from the comments in the script primarily these two:

`// Same script id/url: keep same script`

`// Same script id but url changed: recreate the script`

On those grounds the reason for the ID is only to lookup if the script has been written to page yet. If you are loading up new/different libraries then you want to create a new script tag. Removing old script tags provides no benefits other than human aesthetics normally as the code from the script tags is still running on the page regardless of the presence of the script tag. Querying for existing scripts based on the url is the most effective method to ensure that you don't load things up twice isn't it and should provide all the benefits of using an ID and then comparing the url? Current method isn't exactly causing problems though so if it's not broken don't fix it! 😆

JustFly1984:
can we close PR and according issue?

coulson84:
Not yet. I have just been double checking things and it solves the issue with injectScript but I think there is an issue with useLoadScript which has logic prior to calling to injectScript that causes the useLoadScript hook to complete early.

Will push a fix up in a few minutes.

coulson84:
Okay. Not an issue with useLoadScript hook actually. An issue with my own changes - forgot to add an early return statement in where it and also to remove the old document.getElementById check which meant the promise was still resolving immediately if an element with matching id property was found. Have fixed this and pushed it up.

JustFly1984:
@uriklar @FredyC can you please test this branch before I merge and publish?

coulson84:
@uriklar @FredyC There is a fairly simple reproduction of the bug here
https://codesandbox.io/s/hungry-mountain-r2sc6
(It will crap out the first time it loads or if you press the refresh button on the preview button window)

FredyC:
Well, as I was already mentioning somewhere else, I think makes total sense to get rid of the `id` attribute. Or to be precise I wanted to make it constant, not user-configurable. However, I think that finding an element through the url is indeed much better to avoid duplicates. And removing old ones doesn't matter as those won't be executed again anyway.

I will review the code later, but let's discuss this first. I don't see any benefit of having that `id` attribute. The tomchentw version did not have it either and it was working just fine.

JustFly1984:
@FredyC removing old html elements does make sense, cos we have performance limitations on total amount of html elements on the page 1500-2000, and this line could be crossed without attention

FredyC:
@JustFly1984 Well, as I said, Google adds a bunch of those linked elements which we are not removing. Removing a single script won't do much of good :) You would have to try to identify which scripts were added there by Google and remove those as well, but that's a slippery slope. You might remove a script that wasn't loaded yet.

Anyway, we can basically remove the element right after the `onInit` is called. That script element won't do any good after it's done.

Point is that we surely don't need `id` for a removal.

FredyC:
Thinking about it more, there is a huge limitation in the form of `window.google`. We won't ever be able to overcome it. The latest loaded script will overwrite whatever was there. This is a very important realization and we basically should prevent user trying to load two different configurations of the map.

@slorber Was trying to achieve that in https://github.com/JustFly1984/react-google-maps-api/pull/188#discussion_r305310496, but the approach wasn't right, because it would prevent multiple uses of `useLoadScript` which isn't developer-friendly imo.

I think that probably the simplest and most robust solution would be adding `window.googleUrl` on the first call. Later when we detect an attempt to load maps with different url, we should bail out and inform user it's not possible to do that.

slorber:
@FredyC I don't agree. I want to be able to change script url at runtime, because I need to change the libraries/languages, which requires refetching the script and overriding. It is an SPA, if user changes language, the map should change language at runtime. Your solution could only work for me if you cleanup window.googleUrl on unmount, which imho is not really easy if you allow multiple `useLoadScript` being called at the same time because if there are 2 hooks mounted and one unmounts, you can't do the cleanup yet.

I think it's safe to prevent the user from using `useLoadScript` and don't understand your "developer friendliness". Imagine he does `useLoadScript("fr")` on the header, and `useLoadScript("en")` on the body, what's the purpose of this exactly? Which language should actually be used on the page, as there can only be one at a time? He should rather do just one `useLoadScript("fr")` on a parent top level comp. 

Note that even google logs an error when it gets overriden, there's probably a good reason for it: preventing the user to shoot in the foot. Silently preventing script url changes will just lead to confusion as the user is doing weird things and is not even aware of.

Loading google script is hard and annoying on SPAs, particularly when you have to change the conf dynamically. To ensure it's user friendly imho we should fail fast with meaningful error messages which explain clearly the google map constraints, maybe with a link or something.

JustFly1984:
@FredyC 
I also has requirement to change language in runtime, and agree with @slorber on this topic.

FredyC:
Wait, what? :) So you have `useLoadScript` on top of the app and you dynamically modify `libraries` based on what each page needs? That seems highly ineffective. Why not load all relevant libraries right away?

My use case is simply about the separation of concerns. Most pages don't need maps, so it doesn't make sense to be loading them beforehand. When navigating to the page I want to simply use that hook which has 100% stable configuration across the pages because it's wrapped into another hook which defines it. I certainly don't need a different language or whatever within a single app. Developer friendliness resides in being forced to have that hook once on top and load maps always.

So unless I have misunderstood your use case, I don't see a point why not to prevent changing the configuration the way I've drafted. There wouldn't be any point in having `window.googleUrl` if it would be removed. The reason to prevent loading with different url.

FredyC:
@JustFly1984 
> I also has requirement to change language in runtime, and agree with @slorber on this topic.

Ok so perhaps we should be thinking in a different less-opinionated direction. Provide the most basic implementation that can load the script, but leave the rest to the developer. No limitations or attempts to gracefully take care of dynamic changes. I don't think there is any gold path here to satisfy the needs of everyone.

slorber:
@FredyC 

> Btw, in practice, changing the language on the fly is something that happens once at most per the user. We have opted to simply refresh the app in that case because the hassle of dynamically changing everything is totally not worth it to cover that scenario.

I have an international gatsby site, not all pages use google maps. User can navigate from /fr/pageWithMap to /en/pageWithMap, and the language of google maps must change when this happens. Don't particularly want to refresh the page when this happens, and currently only this lib makes it hard to handle that language change at runtime correctly. If you choose a different path for your app that's fine but do not force everyone to follow your path.

I don't necessarily want to load google maps for all pages, I don't reco to put a global hook for all the site's pages but rather, at the top of each page's body, just under the main site layout, on a per-usecase basis, choose to use the hook or not.

It's the same for libraries. I could load all libraries, but if page1 use lib1, and page2 use lib2, do I really want to load lib1+lib2 if user never visits page2? (actually in this case I do load all the libs because I think the overhead might not be significant)


> Ok so perhaps we should be thinking in a different less-opinionated direction. Provide the most basic implementation that can load the script, but leave the rest to the developer. No limitations or attempts to gracefully take care of dynamic changes. I don't think there is any gold path here to satisfy the needs of everyone.

I've successfully made hacks in my app but those were not easy. On the contrary I think this lib should provide things to solve the pains of this issues. I don't understand why you really need, in your app, to call the hook multiple times, what's the usecase for this? What would be complicated to migrate your multiple calls to the approach I suggest above, which is to put a unique hook somewhere upper in your tree, instead of having duplicate deeper hooks.

As google maps is a singleton, I really feel it's a bad idea to let user call twice the hook on the same app. It's just making things less explicit. As a google maps newbie 1 month ago, I wasn't even aware that google maps was loaded this way and all the subtilities. Had to see only after production that the google maps language did not change correctly on page language change. For me the hook should not fail safe in an unexpected way that lead to production i18n bugs, but rather encourage the developer to do the right thing, not hide him the fact that google maps is a singleton, and educate him on how to handle this annoying singleton properly

JustFly1984:
@FredyC reloading an app is not an option, if you have a delivery form UX, half filled in, for example, and that is just first thing comes to my mind.

The main goal I have made this rewrite is to be able to switch language in runtime.

FredyC:
@JustFly1984 
> reloading an app is not an option, if you have a delivery form UX, half filled in, for example, and that is just first thing comes to my mind.

I would really love to see the user who starts filling the form which is in the wrong language and then he decides to switch language just because it sounds like a good idea :D Doesn't matter if you consider it as a necessary evil, it's understandable. I kinda had this approach before as well, but I see it as micro optimization now.

@slorber 
> I don't understand why you really need, in your app, to call the hook multiple times, what's the usecase for this? What would be complicated to migrate your multiple calls to the approach I suggest above, which is to put a unique hook somewhere upper in your tree, instead of having duplicate deeper hooks.

It's not about calling hook multiple times, I don't do that. I do have a BaseMap component which calls it. So whenever I use that component, I want it to load scripts and be happy. What bothers me is that I would have to include it into some top-level Page component which is totally unrelated to the app and control through the app state if maps should be loaded or not. How is that developer-friendly compared to just using the hook?

slorber:
I have an order form where user can switch country but keep page language. Switching country should trigger map language change because we want Google map addresses to be localized to the order country. This is mostly because in our business we want to support an user with French locale to order for us country and vice versa. 

I understand that you might want to call the hook collocated to the comp that requires Google maps. But this hides the fact there is a singleton. As a newcomer using this lib it can make you think you can display 2 maps in 2 distinct languages at the same time. 

As I said you should absolutely not put the hook at the very top. But you'd rather put it on a per page basis. I'll put this in OrderPage on my usecase because I know Google maps is required on the order process. But if it's too high for you, just find the closest parent of all your comps requiring Google maps. Note that if you only use one map comp per page, you'd still be able to colocate. We may even nooo if you use 2 hooks and they share the same url as it's same. But if you start using 2 hooks with different config, I really think we need fail fast, because weird things will start to happen.

FredyC:
Alright, I gave it some thought and it should probably work. I would even say it clicked in your last comment about the singleton behavior of `window.google` which should be indeed expressed by throwing an error if the developer attempts to have multiple versions. We definitely need to document this well though or people will be lost the same way as I was.

Let's get back to the questions of this PR then.

Should we keep the `id` prop which only complicates things? Changing the url should be enough to take care of loading the correct version. By returning script element reference from loadscript, it should be easy to remove on unmount.

coulson84:
Reading other comments in this thread it sounds like the desired effect, in order to mimic window.google's singleton nature is to only allow one google script tag on page at a time. Therefore it sounds like the script element should remain on page until/if a different google maps configuration is loaded up? Documentation should reflect that changing configuration will cause a refresh of the window.google object to highlight the point raised by @slorber about trying to display two maps in different languages on page at the same time. Also the development build should probably console.warn on configuration change to notify the developer about this as well.

Does the above sound correct to everyone or have I missed anything/made some wrong assumptions? If there is agreement I will update the PR later today.

FredyC:
I think you got it right. Personally, I wouldn't really bother with removing old script element, it's really an micro-optimization, but as long as it's not removed too soon, it should be fine.

Not sure why you want to keep `id` in there at all. Finding script through `url` sounds like a better option as you have suggested.

JustFly1984:
@coulson84 you are correct, there is no possibility to load more than one google maps script, due to Singleton nature of ‘window.google’ object. But there is a clear requirement to change language or libraries, or region e.g url at runtime.

JustFly1984:
Url can change, I’d can’t change

coulson84:
@FredyC removing the old script is for humans more than anything else. So it is more apparent to a dev nosing around the DOM which google maps script/config is active (if they have an app using > 1 config). If only using one script tag at a time then we may as well use an id as we are already. It means the fewest changes to the code is probably marginally quicker than querying by tag+src and also means that if some one is loading their own google maps script, outside of this library for some reason. We will not accidentally target that script.

I'll go ahead and update the PR.

FredyC:
> also means that if some one is loading their own google maps script, outside of this library for some reason. We will not accidentally target that script.

Well, if someone would be that crazy to have `useLoadScript` at one spot and loading maps differently in another spot then I would say they are asking for problems for sure given the singleton nature.

Perhaps the `id` could be there as a sanity check only. We could find element by the url and check `id` to verify it's our element.

I am thinking it's actually a fragile approach to check for the existence of element to decide if we want to inject one or not. What if we really use `window.googleUrl` or something similar to track that state instead? Having a global variable is closer to the singleton approach. Upon the url change, it would be possible to reliably find element by the old url and remove it (with `id` check).

> It means the fewest changes to the code is probably marginally quicker than querying by tag+src

Not sure I got that. Quicker you mean to code that or performance-wise? That would require measurement, but I wouldn't say it's marginally quicker if that's what you meant :)

coulson84:
@FredyC I don't see how using a global variable is any less fragile than checking for the presence of a script element. They are both globally available and open to being manipulated something local to the module code should do the job though

Thinking about this more. The removal of the id property is basically a breaking change and would require a major version update. There's no way of knowing what someone has already implemented on the assumption that the id they pass in is applied to the script and removing that behaviour will break whatever they have implemented so this behaviour needs to remain the same if you want to avoid a major version increase. It might be possible to change the behaviour without a major version hike based on the fact that the changes are to undocumented behaviour that the user should not be relying on anyway... I guess 😕 . Do you see where I am coming from?

FredyC:
It's unlikely anyone would care about `id` prop considering it wasn't present in tomchentw version. Most people probably just migrated without digging deeper.

Either way, I wouldn't be afraid to do a major version. We could even drop the old LoadScript and replace it with LoadScriptNext that utilizes the hook. But in that case, we should have V2 branch and merge a couple of PRs there before merging it to master.

> I don't see how using a global variable is any less fragile than checking for the presence of a script element

It just feels like DOM is something slightly more fragile, not even mentioning it's slower to interact with it than with global variables. Also, we could use the private `Symbol` attached to `window`. Nobody could mess with that.

packages/react-google-maps-api/src/utils/injectscript.ts
```diff
@@ -48,11 +48,11 @@ export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
     script.id = id
     script.async = true
     script.onerror = reject
-    script.setAttribute('state', 'loading')
+    script.setAttribute('data-state', 'loading')
```
FredyC: Do we actually need to set that attribute at this point? Since the only check for `ready` is done and `getAttribute` will return `null`, there is no need to tweak DOM more than necessary imo.

coulson84: Yeah, fair point

coulson84:
Okay. If doing a new major version isn't an issue then I'd like to merge this PR as it stands. I'll do the other work discussed on this thread as a second pull request. 

The reason for splitting these out is that I want the bug fix merging and releasing as soon as possible as I want to use the bug fixed version in our app for work so we push it out in the next release cycle. If all of this work starts to roll in to a new major release then it will probably be delayed by quite a bit.

packages/react-google-maps-api/src/__tests__/utils/injectscript.test.ts
```diff
@@ -0,0 +1,143 @@
+import { injectScript } from '../../utils/injectscript'
+
+interface WindowWithGoogleMap extends Window {
+  initMap?: () => void;
+}
+
+// interface ScriptWithOnErrorHandler extends Element {
```
FredyC: Can you drop these commented out parts unless you want to revive those later?

packages/react-google-maps-api/src/__tests__/utils/injectscript.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      return injectScript({
+        id,
+        url,
+      })
+        .then(value => expect(value).toBeTruthy())
+    })
+
+
+    // it('returns a promise which rejects if the script does not load', () => {
```
FredyC: I wonder why this test doesn't work. Is it some technical limitation?

packages/react-google-maps-api/src/utils/injectscript.ts
```diff
@@ -16,21 +16,30 @@ export const injectScript = ({ url, id }: InjectScriptArg): Promise<any> => {
...
+    const windowWithGoogleMap: WindowWithGoogleMap = window
     if (existingScript) {
       // Same script id/url: keep same script
       if (existingScript.src === url) {
-        return resolve(id)
+        if(existingScript.getAttribute('data-state') === 'ready') {
+          return resolve(id)
+        }
+        else {
+          const onGoogleMapsReadyClosure = windowWithGoogleMap.initMap
```
FredyC: It's kinda hard to understand what's going on here. Had to read it several times to understand it's actually calling original version of that function to resolve all promises in the row. Perhaps even some comment would be useful.
```suggestion
          const originalInitMap = windowWithGoogleMap.initMap
```

packages/react-google-maps-api/src/__tests__/utils/injectscript.test.ts
coulson84: I think it is a limitation with jsdom. It allows setting an onerror handler for the script but always returned a null value when you tried to retrieve it.

coulson84:
@JustFly1984 I don't think the minified script for v1.5.2 has been updated to match the latest unminified content. I can't find any occurrence of `initMap` or `'data-state'` 2 strings which should definitely be in there. They are in the unminified script that gets pulled from npm though. Also if I use the minified version of the script the issue that this PR fixes is not fixed. If I use the unminified 
"reactgooglemapsapi.esm.js" however it has all the relevant changes in and the issue is fixed.

FredyC:
@coulson84 FYI, the latest release should be alright now.

PR #199, opened 2019-08-04
Use load script reset & deprecate ID

FredyC:
Continuation of #188 based on @slorber work.

Instead of hard error when ID prop changes, I've modified it to show a general warning when ID is used as preparation for 2.0 removal of it.

cc @JustFly1984 @uriklar


JustFly1984:
@uriklar can you please test if you have time?

packages/react-google-maps-api/src/LoadScript.tsx
```diff
@@ -27,11 +29,6 @@ export function DefaultLoadingElement() {
   return <div>{`Loading...`}</div>
 }
 
-export const defaultLoadScriptProps = {
```
uriklar: @FredyC why was this extracted to utils? I think it's good co-location wise when default props are in the same place as the component

FredyC: Because of circular dependency, there was a build failure for that. And besides, the `LoadScript` will be most likely removed and replaced with LoadScriptNext in 2.0, so this is a little preparation :)

FredyC:
> can you guys give me some quick bullet points as to what we should be testing here?

Besides deprecation of ID there is now hard error when `useLoadScript` is used more than one at a time. If the previous is unmounted and another mounted, it's ok, but there cannot be two at the time. Then there is a reset for cases when URL would change (eg. different libraries) which did not reset `isLoaded` and `loadError` states before.

packages/react-google-maps-api/src/LoadScript.tsx
JustFly1984: Why LoadScript should be removed? I hear this first time

FredyC: I don't know, it's just assumption considering that `LoadScriptNext` (build on top of `useLoadScript`) is working just fine without all that clean up the mess. All fixes we have been doing lately was in this area. The `LoadScript` just seems to be broken on all fronts.

FredyC:
@JustFly1984 So are you going to merge this? :)

JustFly1984:
@FredyC can you please make an example with gatsby, add docs md and update readme.md?

PR #200, opened 2019-08-05
Fix paths for built files

FredyC:
Discovered another build related issue, we were referring to old files. TSDX has changed the way how are output files named slightly. I've also added cleanup of dist before build to avoid having old code in there.

To explain the ESM format, there is no longer dev/prod variant as that's up to the consumer. It was actually an issue in this project that has lead me to [discover the problem](https://github.com/palmerhq/tsdx/issues/140).

cc @JustFly1984 @uriklar

### DIFF
--- a/packages/react-google-maps-api-marker-clusterer/package.json
+++ b/packages/react-google-maps-api-marker-clusterer/package.json
@@ -21,8 +21,8 @@
   },
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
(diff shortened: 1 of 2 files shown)


FredyC:
@JustFly1984 You might want to publish this soon because a few latest versions are still referring to old ESM build.

JustFly1984:
oh... ok

JustFly1984:
published 1.5.4

PR #204, opened 2019-08-19
Allow for adding a channel to the maps url

codejet:
At the moment it is not possible to add a channel parameter to the maps url by passing it as a prop to `LoadScript`/`useLoadScript`. Channels allow for tracking of usage across different applications using the same client ID. See [Google Maps docs](https://developers.google.com/maps/premium/reports/usage-reports#channels).


codejet:
My pleasure @JustFly1984. Thanks for maintaining this library.

JustFly1984:
@codejet published in 1.6.0

PR #210, opened 2019-08-28
Add InfoBox support

inv8der:
The InfoBox component provides for a lot more customization than InfoWindow, so I thought it would be a cool feature to copy over from tomchentw/react-google-maps. I'm considering switching to this library in a project I'm working on, but the lack of InfoBox support is a blocker.

The implementation is pretty similar to InfoWindow, but it relies on the third-party library `google-maps-infobox`. The InfoBox class this module exports extends `google.maps.OverlayView`. This comes with the unfortunate side-effect that if the global `google` instance changes, all InfoBox instances will no longer render correctly. To get around this issue, I delete from the module cache and re-require `google-maps-infobox` every time the InfoBox component is mounted.


packages/react-google-maps-api/src/components/addons/InfoBox.tsx
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    // google-maps-infobox uses `google` as a global variable. Since we don't
+    // have `google` on the server, we can not use it in server-side rendering.
+    // As a result, we import google-maps-infobox here to prevent an error on
+    // a isomorphic server
+    delete require.cache[require.resolve('google-maps-infobox')]
+    const InfoBox: typeof GoogleMapsInfoBox = require('google-maps-infobox').default
+
+    // @ts-ignore
```
JustFly1984: please do not use //@ts-ignore if possible

JustFly1984: I had similar problem with original marker-clusterer, so I had created new `@react-google-maps/marker-clusterer` copied the code from original, and added types. after that I have added it as dependency into main `@react-google-maps/api`
Can you please copy our marker-clusterer package, and add the code from `google-maps-infobox`, so I will help you to add Typescript to the code, publish it to the npm and so on

inv8der: Sure, I'll try to work on that today

JustFly1984: @inv8der how is your progress?

inv8der: Good. I just need to fix a few more TypeScript errors, then I'll push my changes. I'll do that tomorrow morning. Should I create a new pull request or just use this one?

JustFly1984: use this one. please add yourself to contributors list in package.json

inv8der:
@JustFly1984 I can also add a storybook example if that would be helpful

packages/react-google-maps-api/src/components/addons/InfoBox.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  state: InfoBoxState = {
+    infoBox: null,
+  };
+
+  open = (infoBox: GoogleMapsInfoBox, anchor?: google.maps.MVCObject) => {
+    if (anchor) {
+      infoBox.open(this.context, anchor)
+    } else if (infoBox.getPosition()) {
+      // @ts-ignore
```
JustFly1984: please remove ts-ignore. each ts-ignore hides a bug

PR #218, opened 2019-09-10
Fix broken InfoBox import in docs

emroussel:
# Please explain PR reason.
The example for the `InfoBox` component in docs was broken due to a missing import.

Screen Shot 2019-09-10 at 10 56 01 AM


JustFly1984:
Currently docs deployed from master, but I had no time to publish new version since Infobox was merged. Await for 1.7.0 version this week/

PR #221, opened 2019-09-14
add alt attribute to cluster image

Muskos:
I added `alt` attribute to the cluster.


Muskos:
@JustFly1984 I had a problem after install npm modules in root folder. I couldn't commit while do not remove `node_modules` in root. Does it make sense?

64851199-9560f380-d61f-11e9-9bf1-5768153bebb6

There is Husky pre-commit hook. 
After I remove `node_modules` in root folder there are some warnings:

Screen Shot 2019-09-14 at 8 13 24 PM

Thx you for this app. When you will create a new version with this fix?

PR #222, opened 2019-09-14
first attemplt to implement hooks

JustFly1984:
# Please explain PR reason.
I've started to work on react hooks, anybody interested can join, or give advice from where I left it today.

### DIFF
--- a/packages/react-google-maps-api/src/GoogleMapNext.tsx
+++ b/packages/react-google-maps-api/src/GoogleMapNext.tsx
@@ -1,0 +1,287 @@
+import * as React from "react"
+
+import MapContext from "./map-context"
+
+import {
+  unregisterEvents,
+  applyUpdatersToPropsAndRegisterEvents
+} from "./utils/helper"
+
+const eventMap = {
+  onDblClick: "dblclick",
+  onDragEnd: "dragend",
+  onDragStart: "dragstart",
+  onMapTypeIdChanged: "maptypeid_changed",
+  onMouseMove: "mousemove",
+  onMouseOut: "mouseout",
+  onMouseOver: "mouseover",
+  onMouseDown: "mousedown",
+  onMouseUp: "mouseup",
+  onRightClick: "rightclick",
+  onTilesLoaded: "tilesloaded",
+  onBoundsChanged: "bounds_changed",
+  onCenterChanged: "center_changed",
+  onClick: "click",
+  onDrag: "drag",
+  onHeadingChanged: "heading_changed",
+  onIdle: "idle",
+  onProjectionChanged: "projection_changed",
+  onResize: "resize",
+  onTiltChanged: "tilt_changed",
+  onZoomChanged: "zoom_changed"
+}
+
+const updaterMap = {
+  extraMapTypes(map: google.maps.Map, extra: google.maps.MapType[]) {
+    extra.forEach(function forEachExtra(it, i) {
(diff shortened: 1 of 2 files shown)


JustFly1984:
@uriklar can you please look?

packages/react-google-maps-api/src/GoogleMapNext.tsx
```diff
@@ -0,0 +1,287 @@
...
+}
+
+export function GoogleMap(props: GoogleMapProps): JSX.Element {
+  const [map, setMap] = React.useState<google.maps.Map | null>(null)
+
+  const [registeredEvents, setRedirectEvents] = React.useState<google.maps.MapsEventListener[]>([])
+
+  const mapRef = React.useRef<HTMLDivElement>(null)
+
+  React.useEffect(function onMount () {
```
uriklar: Why do we need to separate mount and update to two separate useEffects functions?
This is a bit counter intuitive to the hooks way of things..

I think they could be united to one useEffect that maybe has a condition to handle mount specific logic

packages/react-google-maps-api/src/GoogleMapNext.tsx
```diff
@@ -0,0 +1,287 @@
...
+  onTiltChanged?: () => void
+  /** This event is fired when the map zoom property changes. */
+  onZoomChanged?: () => void
+  /** This callback is called when the component unmounts. It is called with the map instance. */
+  onLoad?: (map: google.maps.Map) => void | Promise<void>
+  /** This callback is called when the map instance has loaded. It is called with the map instance. */
+  onUnmount?: (map: google.maps.Map) => void | Promise<void>
+}
+
+export function GoogleMap(props: GoogleMapProps): JSX.Element {
```
uriklar: I'm not sure if the GoogleMap component follows the exact same logic as the others, but when I initially tried implementing hooks the idea was creating a `useMapComponent` hook that accepts the component type, it's eventMap and updaterMap. 

Check out (not a working implementation, but the idea I think is pretty much there):
The "generic" hook:
https://github.com/JustFly1984/react-google-maps-api/blob/1b1940eef9a293c7e4743484bb75046e65bd8d5b/packages/react-google-maps-api/src/utils/use-map-component.ts

And a component that uses it:
https://github.com/JustFly1984/react-google-maps-api/blob/1b1940eef9a293c7e4743484bb75046e65bd8d5b/packages/react-google-maps-api/src/components/drawing/Rectangle.tsx

packages/react-google-maps-api/src/GoogleMapNext.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+      registeredEvents = applyUpdatersToPropsAndRegisterEvents({
+        updaterMap,
+        eventMap,
+        prevProps,
+        nextProps: props,
+        instance: map
+      })
+    }
+  }, [props])
```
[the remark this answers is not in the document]
vvmarulin: And the same approach with `useRef` have to be used for caching of `registeredEvents`. React state has to be used to keep value that will be rendered. Saving `registeredEvents` in a state will make component renders one extra unnecessary time.

packages/react-google-maps-api/src/GoogleMapNext.tsx
lluisrojass: @uriklar curious, any reason this example omits the onLoad prop ? it's not due to any limitation is it?

PR #228, opened 2019-09-18
Update Data.tsx

danloiterton:
Data component events incorrectly typed with an argument of `google.maps.MouseEvent`. They should instead be `google.maps.Data.MouseEvent`.
See https://developers.google.com/maps/documentation/javascript/reference/data#Data

(Awesome library by the way!!)


JustFly1984:
published 1.7.6

enjoy!

danloiterton:
Fantastic! Thanks for that. I'll add myself next time I make a fix 👍

PR #231, opened 2019-10-01
Fix using 'TOP_LEFT' as controlPosition in Data docs

emroussel:
# Please explain PR reason.

Using the string `'TOP_LEFT'` as the value for `controlPosition` of `Data` produces this error: `InvalidValueError: setControlPosition: TOP_LEFT`.

Changed that to just use the actual value from google maps, but had to wrap it in a condition because `window.google` started out as `undefined`.


uriklar:
Thanks!

PR #249, opened 2019-10-30
fix(Marker): correct options type

FallOutChonny:
It should be `google.maps.MarkerOptions` not `google.maps.MapOptions`


FallOutChonny:
Thanks for merging!

PR #259, opened 2019-11-06
Fix OverlayView props ts definition

### LINKED ISSUE #257: Typescript type for OverlayView position is wrong

## Your Environment

@react-google-maps/api version ^1.7.8

## How is it behave?

Typescript type for OverlayView position is wrong. It is specified to be `google.maps.LatLng` but should be `google.maps.LatLng | google.maps.LatLngLiteral`. This means that while this prop accepts { lat: ..., lng: ... } as position at runtime it cannot be used with typescript.

Typescript types should specify that it accepts LatLngLiteral

CodeWitchBella:
Fixes #257


JustFly1984:
published in 1.7.9

PR #264, opened 2019-11-18
after logging the error from inject script rethrow it

coulson84:
I realised that I forgot to put this important line into my previous commit. Sorry for making this be split over two releases now 🙈 

After logging the error from inject script ensure that we rethrow the original error so that any promises chain outside of the `injectScript` function is still able to catch the error and handle it according to their own logic. Specifically used in useLoadScript hook. Which, without this fix always assumes that the script loads and then error when trying to access `window.google.maps`


JustFly1984:
published 1.7.12

PR #414, opened 2020-01-17
Documentation for two functions was mixed up

lukahartwig:
I saw that the documentation for two of the GoogleMap props was mixed up.


JustFly1984:
@lukahartwig thank you for contribution!

PR #870, opened 2020-04-24
docs(api): add needed props to get a functional rendered map

StanleySathler:
Great job, buddies! :star2: It's been hard to find a maintained library for integrating Google Maps into a React application.

I took the opportunity to add a few changes in the _Getting Started_ section.

Without these properties, we can't see either anything on the screen nor any error message. It might confuse both beginners and even experienced developers.

This PR helps to add the basic props in order to get a functional rendered map on the screen. The users can start working from this point.

In order to have the simplest form, I also removed the `id` props as they seem to be not really required (at least not to get a functional map).

### DIFF
--- a/packages/react-google-maps-api/src/docs/getting-started.md
+++ b/packages/react-google-maps-api/src/docs/getting-started.md
@@ -13,6 +13,8 @@ The two basic components required to use this library are:
 * LoadScript - Loads the Google Maps API script
 * GoogleMap - The map component inside which all other components render
 
+The simplest way to get a functional map is:
+
 ```js static
 import React, { Component } from 'react'
 import { GoogleMap, LoadScript } from '@react-google-maps/api'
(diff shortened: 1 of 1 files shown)


JustFly1984:
@StanleySathler  Thank you for your contribution to our project! 

Can you please change code, so there is no new objects or new arrays passed to the props directly? 

Passing new objects and arrays directly, causes re-renders and harm performance, which adds up to confusion among new comers a lot. I'm talking about `mapContainerStyle`, `center` props specifically. Also we could add the comment that this props should be cached in state, React.useMemo or const outside of component function scope.

JustFly1984:
The second idea is to provide 2 minimal examples - first with `React.Component` as you improved and second as functional component, wrapped in React.memo()

StanleySathler:
Sure, @JustFly1984! I'll handle both soon! :rocket:

StanleySathler:
@JustFly1984 most recent changes were pushed (I did a force-push due a rebase - let me know if you prefer normal merging commits instead).

Is this the idea you originally thought? Suggestions are welcome :rocket:

Issue #10, opened 2018-12-01
Allow InfowWindow anchor to be a <Marker/> component


uriklar:
Right now supports only position property

grdpr:
Hi, I'm trying to make a Circle marker with a InfoWindow popup. Any documentation or example about it? Many thanks in advance

uriklar:
Hi @grdpr, 
Something like this? https://codesandbox.io/s/2vx96zljk0
You can give the marker a circle icon using the icon property, or even use an actual circle instead

grdpr:
Thanks @uriklar! I'll give a try to render a list of markers with this InfoWindow and also with OverlayView to feel the difference. OverlayView isn't include in the API from GoogleMaps right?

uriklar:
Included indeed: https://developers.google.com/maps/documentation/javascript/reference/overlay-view#OverlayView
Take note that you don't need to handle the onAdd, onDraw etc... callbacks. We handle them for you.

grdpr:
Thanks for all this work! 
I'm struggling by testing many react-google-map-package. 
Is there any onClick event possible on the <GoogleMap /> wrapper to listen for a click?

uriklar:
Not sure what you mean by "wrapper" but almost every component exposes an onClick event handler

grdpr:
@uriklar the onClick event handler works well on the map! 
But I'm still struggling to render multiple markers with InfoWindow. When I open one infoWindow, all infoWindow are opening.

urikphytech:
Do you want to share a codesandbox?
I wouldn't create multiple info windows. I'd create one info window and pass it the props depending on the selected marker (I.E - the marker's position and any info related to the selected marker you want to show)

grdpr:
here it is: https://codesandbox.io/s/l2jn4xm9yl

michaelsync:
@grdpr  https://codesandbox.io/s/l2jn4xm9yl returns 404.

tdotholla:
how do we pass the anchor `Marker` object from an `onHover`? 
seems like we can only populate some state or store with `onLoad`

uriklar:
@tdotholla I didn't quite understand what you're asking

tdotholla:
> @tdotholla I didn't quite understand what you're asking

No problem. it was explained better by @FredyC here (https://github.com/JustFly1984/react-google-maps-api/issues/18#issuecomment-481849486) 

[...]

PR #879, opened 2020-05-12
Anchoring InfoWindow children of Markers

joshcoales:
# Please explain PR reason.
This is to resolve https://github.com/JustFly1984/react-google-maps-api/issues/10

It allows putting an InfoWindow as a child of a Marker, and having the infowindow automatically anchored to the marker.

### DIFF
--- a/packages/react-google-maps-api/src/components/drawing/Marker.tsx
+++ b/packages/react-google-maps-api/src/components/drawing/Marker.tsx
@@ -249,11 +249,20 @@ export class Marker extends React.PureComponent<MarkerProps, MarkerState> {
   }
 
   render(): React.ReactNode {
-    let children: ReactNode[]|null = null
+    interface HasMarkerAnchor {
+      anchor?: google.maps.Marker | null
+    }
+
+    let children: ReactNode | null = null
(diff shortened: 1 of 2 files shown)


joshcoales:
Here's an example of it working, in a gist:
https://gist.github.com/joshcoales/d89a2f947b9c3b3c4ddafaebfca1df6b
The imports are messy so I could get it working locally.. There's probably a cleaner way!

packages/react-google-maps-api/src/components/drawing/Marker.tsx
```diff
@@ -248,7 +249,22 @@ export class Marker extends React.PureComponent<MarkerProps, MarkerState> {
   }
 
   render(): React.ReactNode {
-    return this.props.children || null
+    interface HasMarkerAnchor {
+      anchor?: google.maps.Marker | null
+    }
```
JustFly1984, line 254: please move interface to types file, and import it into this file

joshcoales:
There we go, pulled out that interface in 6fddef0eee75da462336f36c1e3092b63ab4a304

Issue #49, opened 2019-02-16
Reference Map object from outside the GoogleMap component


buzzguy:
Thanks for the quick response to my last issue.

I'm able to control the map object from within the onLoad function, but I wanted to have an airbnb like effect of hovering over a place card which would trigger a panTo that marker on the map and perhaps a toggle of an infowindow. 

would you know the best way of doing that?

thanks

uriklar:
Hi @buzzguy. You say you have the map object from onLoad, why not save it somewhere (class variable for instance) and then you'll be able to control it whenever you wish?

FredyC:
It might easier and more robust to use a Context for that purpose. Perhaps even on the side of the library, it would make sense except it could be `undefined` till loaded, but that should be fine.

uriklar:
We use context to pass the map down to it's children, but @buzzguy is talking about accessing the map from the parent that renders it

buzzguy:
@uriklar  Hi, I do have the map object that exists within the onLoad. 
The hierarchy of the components is basically:
- Page
    - List of places component
	    - place item
    - Map Component
	    - markers

So if I want to give place item scope to the map object inside of Map Component, I should save it as a class variable? Would you have an example?

I have the map object from within the onLoad method of GoogleMap from the example (below).
```
callBack = map => {
    const bounds = new window.google.maps.LatLngBounds();
    map.fitBounds(bounds);
  }
render(){
<GoogleMap
  onLoad={this.callBack}
/>
```

uriklar:
First of all, this question isn't really react-google-maps-api related.. Second, I can try and help, but i'd have to see some code. Can you link a codesanbox with what you're trying to do?
In general, once you get the map from onLoad you can save it wherever you want, be it a class variable, state, context...

buzzguy:
@uriklar right on, you're right. I figured it out. I just set this.map = map in the callBack. thanks!

PR #880, opened 2020-05-13
Export map context

Kjagd:
By exporting the `MapContext` it is possible to lift the context of the map up, allowing components to render map components without being a child of the `GoogleMap` component.

```tsx
...
const [myMap, setMyMap] = useState(null)

return (
  <Content>
    <GoogleMap onLoad={setMyMap} />
    <MapContext.provider value={myMap}>
      <Sidebar>
        <MyMarker />
      </SideBar>
    </MapContext>
  </Content>
)
...
```

This is currently very clunky to do, as you would have to save the map instance in a state and pass your own context around. But since the map components are looking for the `MapContext` by default, you would have to do something like

```tsx
<Marker onLoad={m => map && m.setMap(map)} />
```
This was discussed a bit here as well: https://github.com/JustFly1984/react-google-maps-api/issues/49

An alternative to this could also be to allow passing in a custom `ref` to `GoogleMap`, allowing us to mount the map itself on a child element...

### DIFF
--- a/packages/react-google-maps-api/src/index.ts
+++ b/packages/react-google-maps-api/src/index.ts
@@ -75,4 +75,4 @@ export {
 
 export { default as Autocomplete, AutocompleteProps } from './components/places/Autocomplete'
 
-export { useGoogleMap } from './map-context'
+export { default as MapContext, useGoogleMap } from './map-context'


Kjagd:
@JustFly1984 Any thoughts on this? I've been using it with much success for a week now.

PR #917, opened 2020-06-04
Added `MarkerWithLabel` component (based on `markerwithlabel`).

### LINKED ISSUE #219: Addition of MarkerWithLabel component
We're migrating away from `react-google-maps` and hit the roadblock of not having the [`MarkerWithLabel`](https://tomchentw.github.io/react-google-maps/#markerwithlabel) support. Is this on the roadmap? Didn't see it mentioned in #18 

Thanks!

### LINKED ISSUE #782: Integrate markerwithlabel to extend Marker functionality
Hi there,

So i've needed to display labels on my markers and the standard <Marker> components only allows for the standard label provided by google maps. It could be nice to integrate the functionality of **markerwithlabel** found [here](https://www.npmjs.com/package/markerwithlabel).

Theoretically, everything is the same as the Marker class except with additional label parameters and should almost just be a copy of the <Marker> component.

For now, I'll just use the project directly but I thought I'll post here to maybe see that added to the library.

BowlingX:
This PR adds the `MarkerWithLabel` component that allows you to create a custom marker from a react-component. Fixes #782, fixes #219

I'm unable to run the linter locally as it seems like prettier is missing in the devDependencies, or it is suppose to be done differently?

I decided to extend the current Marker component instead of extending it's current functionality to give the user the choice to include the `markerwithlabel` dependency or not. I think in this case inheritance makes sense.

Thank you


JustFly1984:
@BowlingX  Sorry, but this library does not use external dependencies, especially which does not support typescript.. You can fork markerwithlabel, and place it in packages/react-google-maps-api-marker-with-label and reuse tsdx setup from other packages to refactor it to typescript, and after that I will publish it and include as dependency to main project. Please continue in same PR.

BowlingX:
Thank you for having a look :). Rewriting the library / Adding types for it is out of my scope currently timewise, but I can provide types for it, as it basically just extends `google.maps.Marker` and takes some more properties (extending `google.maps.MarkerOptions`). Let me know if this can work as well.

BowlingX:
I also tried to use the current version directly from google (`@google/markerwithlabel`), but it seems there is a bug on the latest version (see https://github.com/googlemaps/v3-utility-library/issues/651)

JustFly1984:
@BowlingX well, if you clone markerclusterer to packages/react-google-maps-api-marker-with-label, I will help to refactor it to typescript and publish

JustFly1984:
@BowlingX I've spent a weekend to give some love to the project. I had rebased your PR to 2.0.0 version branch. I have created new package marker-with-label, but had no time to refactor it to TS yet. Will continue then I will have time again. BTW I've published 2.0.0-alpha and would love if you test it.

JustFly1984:
PS please fix merge conflicts

BowlingX:
I can't test the implementation currently as it seems like there are several issues:

```
VM762:11 Uncaught SyntaxError: Identifier 'InfoBox' has already been declared
    at new Function (<anonymous>)
```

...and something is triggering and endless loop:

```
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    in ScriptLoaded (created by bound anonymous)
    in bound anonymous (created by ReactExample)
    in Wrapper (created by ReactExample)
    in ReactExample
```

packages/react-google-maps-api/src/components/drawing/MarkerWithLabel.tsx
```diff
@@ -33,7 +34,13 @@ const markerWithLabelUpdaterMap = {
   ...updaterMap,
 }
 
-const MarkerWithLabel: FC<MarkerWithLabelProps> = props => {
+const MarkerWithLabel: FC<MarkerWithLabelProps> = ({
+  labelAnchor,
```
BowlingX, line 37: Something also seems to be wrong with the eslint setup, I get warnings where I should not get them.

JustFly1984: can you attach a screenshot of warnings?

I will continue to work on alpha version this week, tried alpha in codesandbox example and it fails.

BowlingX:
I have a general  question, why not using the new @google packages for the implementations (e.g. @google/markerclustererplus etc.) instead of porting /copieng the dependencies. It seems redundant to me and will make maintainability harder espacially when those are maintained by google directly against there map API.

JustFly1984:
@BowlingX if it does not support types, it is buggy. It is still buggy as it is. It is not possible to be involved in maintenance of google packages. General quality is sucks. I'm trying to improve general dev experience, and use it as learning experience for myself.

rsanath:
Thanks guys. any idea when this will get merged?

JustFly1984:
at this point this PR requires cherrypick to another PR to development branch. Sorry I just have no time to do that. If somebody want to be a champion, I will review and merge, and release new version.

PR #929, opened 2020-06-07
refactor to React hooks

JustFly1984:
# Please explain PR reason


packages/react-google-maps-api/src/components/drawing/Circle.tsx
```diff
@@ -96,71 +98,69 @@ export interface CircleProps {
...
+          instance,
+        }
+      )
+
+      return function cleanup(): void {
+        unregisterEvents(registeredEvents)
+      }
+    },
+    [props, instance, prevProps]
+  )
```
uriklar, line 161: Can we extract this to a custom hook used by all components? (Or maybe just the one's that have this exact functionality)

What i'm concerned about is that any time we need to make a change to this logic we'll have to change it in many components

packages/react-google-maps-api/src/components/drawing/Circle.tsx
```diff
@@ -96,71 +98,69 @@ export interface CircleProps {
...
+  const prevProps: CircleProps = usePrevious<CircleProps>(props)
+
+  const [instance, setInstance] = React.useState<google.maps.Circle | null>(
+    null
+  )
+
+  React.useEffect(
+    function effect() {
+      if (map !== null) {
+        if (instance === null) {
```
uriklar, line 113: These nested if statements are a bit hard to read. I'd try going the other way around. 
Start with the conditions where this effect does nothing and add an early return. 

Something like (psuedo code..):
if (!map) { return;}

if (!instance) {
  setInstance(...)
}

packages/react-google-maps-api/src/components/drawing/Circle.tsx
```diff
@@ -96,71 +98,69 @@ export interface CircleProps {
...
+          )
+        }
+
+        if (instance !== null) {
+          instance.setMap(map)
+
+          if (onLoad) {
+            onLoad(instance)
+          }
+        }
```
uriklar, line 128: If I understand correctly this if will only be triggered after an additional render cycle caused by setInstance. 
Maybe we can do all the work in one render cycle by caching the instance result in a variable

```
const _instance = setInstance(...)
_instance.setMap(map)
if (onLoad) { onLoad(_instance)}
```

packages/react-google-maps-api/src/components/drawing/Circle.tsx
uriklar, line 113: Maybe we can extract this also to a custom hook? It would need to receive a generic to indicate the instance type, but i think instance, setInstance and this useEffect can be in a generic useMapComponent hook

PR #996, opened 2020-06-19
Accessibility: add empty alt text to 'x' icon image in InfoWindow

vanderhoop:
This one-line change addresses an accessibility violation for screen readers, in that the image used for the 'x' icon in the `InfoWindow` lacks alt text. 

The alt is now an empty string due to the image being decorative rather than informative, as the image is already wrapped in a `<button>` with `aria-label="Close"`


JustFly1984:
@vanderhoop can you please also set aria-hidden='true' on img?

vanderhoop:
@JustFly1984 added `aria-hidden="true"` in adf700e

PR #1018, opened 2020-06-22
fix: allow loadScript to check for window.google.maps intead of windo…

byrekt:
…w.google

# Please explain PR reason
Google's One Tap sign-in uses the `window.google` object to store account information. This library does not work when using One Tap because it thinks that the library is already loaded since this library only checks for `window.google`. So I've made a modification to check for `window.google.maps` instead.


byrekt:
@JustFly1984 can you review this please?

JustFly1984:
released 1.9.8

PR #1046, opened 2020-06-29
amend docs for MarkerClusterer

capaj:
also fix formating for files that weren't formatted with prettier


JustFly1984:
released 1.9.8

PR #1145, opened 2020-07-22
Make Autocomplete.tsx stylable

### LINKED ISSUE #201: allow to style autocomplete
Right now it's not possible to style the[ wrapping div.](https://github.com/JustFly1984/react-google-maps-api/blob/master/packages/react-google-maps-api/src/components/places/Autocomplete.tsx#L136) I believe just supporting style/className ought to be enough.

@react-google-maps/api@1.5.4

schoenwaldnils:
Currently the autocomplete ref is a div that is not styleable.
This changes should add this option.

Were not able to set up locally. Please test it.

Fixes #201


JustFly1984:
@schoenwaldnils spreading props on html element could lead to props leaks in html.
Also using any in typescript projects is not safe.

Please create className prop with string type instead.

JustFly1984:
@schoenwaldnils please test 1.9.12

PR #1443, opened 2020-09-28
Support map ids for vector maps

Kjagd:
Google now supports using vector maps instead of raster (in beta), see https://developers.google.com/maps/documentation/javascript/vector-map :tada: 

In order to support this, the load script url needs to include the available map ids.

Additionally, the `MapOptions` should include the specific `mapId`, however the official types are not yet updated, as evident from their own examples [here](https://github.com/googlemaps/js-samples/blob/25a4ff9635382931e19cb723b3c9460076af1d0e/samples/vector-use-static-map/src/index.ts#L43).

However, since all the option props from the `GoogleMap` component are applied, we can just use
```ts
<GoogleMap
  onLoad={onMapMount}
  options={{ mapId: '...' } as google.maps.MapOptions }
/>
```
and wait for the official types to get updated.

I've tested it, and it works pretty well. 

image

### DIFF
--- a/packages/react-google-maps-api/src/useLoadScript.tsx
+++ b/packages/react-google-maps-api/src/useLoadScript.tsx
@@ -28,6 +28,7 @@ export function useLoadScript({
   libraries,
   preventGoogleFontsLoading,
   channel,
+  mapIds,
 }: UseLoadScriptOptions): {
   isLoaded: boolean
   loadError: Error | undefined
@@ -73,6 +74,7 @@ export function useLoadScript({
     region,
     libraries,
     channel,
+    mapIds
   })
 
   React.useEffect(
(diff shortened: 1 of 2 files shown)


JustFly1984 (approved):

JustFly1984:
released 1.11.0

jensfranke:
@Kjagd / @JustFly1984  Sadly, from my point of view it looks wrong or did google changed it?

It should be `mapId` (your example) and not `mapIds` (your code) and it should be and option for `mapOptions`:

https://developers.google.com/maps/documentation/javascript/webgl?hl=de#vector-id
https://developers.google.com/maps/documentation/javascript/reference/map#Map.setMapTypeId

Please correct me if I'm wrong.

PR #1786, opened 2020-12-07
changed default version for load script to quarterly #1776

robert52:
For a more stable map loading we could change the default **weekly** to a more predictable version like **quarterly**

> The quarterly channel is specified with v=quarterly.
> This channel is updated once per quarter, and is the most predictable.

[Source](https://developers.google.com/maps/documentation/javascript/versions)


JustFly1984:
I think weekly is more appropriate, you can set any

PR #2573, opened 2021-05-21
Fix CALCULATOR in Clusterer

KrzysztofKarol:
### BREAKING CHANGE

Documentation says that for 125 markers index 3 is returned
> The styles array element used has an index of index minus 1. For example, the default calculator returns a text value of "125" and an index of **3** for a cluster icon representing 125 markers so the element used in the styles array is 2.

Unfortunately current code returns index 4

Code:
```javascript
const CALCULATOR = function CALCULATOR(
  markers: MarkerExtended[],
  numStyles: number
): ClusterIconInfo {
  let index = 0

  const title = ''

  const count = markers.length.toString()

  let dv: string | number = count

  while (dv !== 0) {
    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
    // @ts-ignore
    dv = parseInt(dv, 10) / 10

    index++
  }

  index = Math.min(index, numStyles)

  return {
    text: count,
    index: index,
    title: title,
  }
}
```

```javascript
CALCULATOR(Array.from({ length: 125 }), 10)
// {text: "125", index: 4, title: ""}

### DIFF
--- a/packages/react-google-maps-api-marker-clusterer/jest.config.js
+++ b/packages/react-google-maps-api-marker-clusterer/jest.config.js
@@ -1,0 +1,18 @@
+module.exports = {
+  "roots": [
+    "<rootDir>/src"
+  ],
+  "transform": {
+    "^.+\\.tsx?$": "ts-jest"
+  },
+  "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.tsx?$",
+  "moduleFileExtensions": [
+    "ts",
+    "tsx",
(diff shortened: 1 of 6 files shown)


JustFly1984 (approved):
Thank you! I will release a new version this week

JustFly1984:
@KrzysztofKarol I've released new version with your fix. please test in your App. 
I have issues with updating Gatsby examples.
Docs deployed without issues.

PR #2771, opened 2021-07-14
Add mapId option

zakjan:
`mapId` option enables vector maps, tilt and rotation
https://developers.google.com/maps/documentation/javascript/webgl
https://developers.google.com/maps/documentation/javascript/webgl/tilt-rotation


Kjagd:
hi @zakjan, I've previously added support here: https://github.com/JustFly1984/react-google-maps-api/pull/1443
The current issue is that the google map types are not updated. However, I think the mapId should be included in `options` prop (which uses the DT google maps types) and not added here.

zakjan:
Thanks, I found it later already that it's supported in options, so I use it there.

There are many properties that are supported both standalone and in options, I'm not sure what's the difference.

Issue #2792, opened 2021-07-21
How to use custom svg marker without http url ?


Aarbel:
Cf in the docs https://github.com/JustFly1984/react-google-maps-api/blob/7e807a6b9115766609c667bef783eb01f706f4c7/packages/react-google-maps-api/src/components/drawing/Marker.md#custom-marker-icon-example

![image](https://user-images.githubusercontent.com/25119847/126536957-aceebe29-e417-41be-bce7-a5418b4e4f74.png)

Here you provide an example with a custom url png marker, but how should be the code for a custom svg element coming from React with google api v3 ? (and not given by a server / url, or only compatible with old v1 or v2 api)

I checked on stackoverflow, i saw dataUri string conversion, etc... but nobody seems to have a definitive / stable solution.

# Links

https://github.com/tomchentw/react-google-maps/issues/563
https://medium.com/@derekbrimley/dynamic-custom-google-map-markers-66b4c22d13e
https://stackoverflow.com/questions/24413766/how-to-use-svg-markers-in-google-maps-api-v3

Aarbel:
Cf https://github.com/JustFly1984/react-google-maps-api/issues/225

Aarbel:
Maybe the beginning of a solution: https://medium.com/@ians/rendering-svgs-as-images-directly-in-react-a26615c45770

Aarbel:
Using this lib can also help reducing svg size: https://github.com/tigt/mini-svg-data-uri

Aarbel:
Ok problem solved, just created this PR: https://github.com/JustFly1984/react-google-maps-api/pull/2794

PR #2794, opened 2021-07-22
Add pure React example for marker rendering

Aarbel:
# Please explain PR reason

Added an example for markers creation, without raw url. 

This example is here to show that markers can be created with React, then rendered in Google Maps

closes https://github.com/JustFly1984/react-google-maps-api/issues/2792


JustFly1984 (approved):
Amazing!

PR #2852, opened 2021-09-15
WIP: Using Rollup to bundle the library

thekevinbrown:
- Switched from tsdx to rollup for bundling the main library.
- Rollup emits CJS, ESM, and UMD output with sourcemaps and a .d.ts file for the module.
- Rollup also emits a minified version of each of the above for convenience.
- Added a .npmignore file to avoid publishing the src/ directory in the final NPM publish.

I've applied this to:
- `react-google-maps-api`
- `react-google-maps-api-infobox`
- `react-google-maps-api-marker-clusterer`

I didn't apply this to `react-google-maps-api-marker-with-label` as it looks like there's not much in that package. Let me know if you like the approach, and if there's anything else that needs doing to unblock releases I'm happy to help.


JustFly1984 (approved):

gkrinc:
@thekevinbrown I'm still see CSP `unsafe-eval` errors in the console after updating to v2.3.0. Is there particular APIs I need to avoid? I'm only using `GoogleMap` and  `Marker` components.

Issue #154, opened 2019-06-11
make `create-react-app` example in packages, same way as gatsby-example


JustFly1984:
I think we can externalize common example components to separate packages, and reuse them in both, gatsby and create-react-app examples, as well as other people can reuse them in their own apps.

PR #2892, opened 2021-10-16
Google marker clusterer

mradenovic:
[@googlemaps/markerclusterer](https://github.com/googlemaps/js-markerclusterer) is an official Google library for marker clusters. According to its home page, it is a continuation of the library in this project. It looks pretty active and well maintained. Adds some new and exciting features to clusters.

It can be used as an option in this project, and the old one is kept for backward compatibility, or it can completely replace the old `MarkerClusterer` at some point.

### DIFF
--- a/packages/react-google-maps-api/package.json
+++ b/packages/react-google-maps-api/package.json
@@ -93,6 +93,7 @@
   },
   "dependencies": {
     "@googlemaps/js-api-loader": "1.12.5",
+    "@googlemaps/markerclusterer": "^1.0.8",
     "@react-google-maps/infobox": "2.5.0",
     "@react-google-maps/marker-clusterer": "2.5.0",
     "@types/google.maps": "3.45.6",
--- a/packages/react-google-maps-api/package.json
+++ b/packages/react-google-maps-api/package.json
@@ -93,7 +93,7 @@
   },
   "dependencies": {
     "@googlemaps/js-api-loader": "1.12.5",
-    "@googlemaps/markerclusterer": "^1.0.8",
+    "@googlemaps/markerclusterer": "1.0.8",
     "@react-google-maps/infobox": "2.5.0",
     "@react-google-maps/marker-clusterer": "2.5.0",
     "@types/google.maps": "3.45.6",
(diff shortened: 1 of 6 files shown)


JustFly1984 (approved):

JustFly1984:
@mradenovic I have released 2.6.0, please test. Also we have issues with gatsby 4 upgrade, of somebody want to fix gatsby example, PR is welcome.

mradenovic:
> @mradenovic I have released 2.6.0, please test. Also we have issues with gatsby 4 upgrade, of somebody want to fix gatsby example, PR is welcome.

@JustFly1984 I tried `yarn build:gatsby` on fresh clone and I had no issues. BTW, what is the purpose of gatsby example?

- If it is for development as described [here](https://github.com/JustFly1984/react-google-maps-api/tree/develop/packages/react-google-maps-api#how-to-test-changes-locally), I find storybook much easier to use.
- If it is for display at [demo page](https://react-google-maps-api-gatsby-demo.netlify.app/), it need some love.
- If it is to show example code, the whole gatsby example obscures the examples and something like #154 should be done.

Should we open a new discussion or an issue for this?

JustFly1984:
It is a demo implementation with a bunch of examples. It does deploy. Docs are broken and outdated.
Gatsby v4 break for me on dev.

mradenovic:
> Gatsby v4 break for me on dev.

What exactly doesn't work?

everdrone:
I'm getting a lot of warnings `google.maps.event.addDomListener() is deprecated, use addListener instead` when using `MarkerClusterer` from this library., until the pages freezes completely

JustFly1984:
@everdrone  Looks like you have a memory leak and perf issues.
Please use eslint with react react-perf and react-hooks plugins. Also use `memo` on exports to minimize re-renders.

If you want the code change, please provide a PR, and I will merge it and release new version. Currently there is https://github.com/JustFly1984/react-google-maps-api/pull/2989 PR work in progress to support react@18, and I've started to refactor PureComponents into functional. This looks like requirement to rewrite the module. current deprecated method can't be just renamed.

everdrone:
I'll have a more in depth look at the problem, I also found that using the [method described here](https://github.com/tomchentw/react-google-maps/issues/836#issuecomment-894381349) the amount of warnings reduces drastically (from ~19k to 1k per render).

Seems like the problem is originating from within the clustering api itself (using the deprecated method). Also because this problem occurs only when the browser developer tools are open. So it probably has to do with how many warnings it can throw in the console

Issue #2981, opened 2022-04-05
Tries to get Google Fonts even when `preventGoogleFontsLoading` is set


polys:
# Issue template

If you want to ask question, ask it in Github Discussions, [Spectrum.chat](https://spectrum.chat/react-google-maps) or [Slack channel](https://join.slack.com/t/react-google-maps-api/shared_invite/enQtODc5ODU1NTY5MzQ4LTBiNTYzZmY1YmVjYzJhZThkMGU0YzUwZjJkNGJmYjk4YjQyYjZhMDk2YThlZGEzNDc0M2RhNjBmMWE4ZTJiMjQ)

Please do not post unformatted code into issues, and please do not ask questions. Only real issues, PR or feature requests allowed. Minimal reproduction in codesandbox.io is must have.

## Your Environment

OS: Windows

Node 16.14.2

React 17.0.2

@react-google-maps/api 2.8.1

## How is it behave?
It makes requests to https://fonts.gstatic.com and https://fonts.googleapis.com

## How should it behave correctly?
It shouldn't make those requests.

JustFly1984:
@polys your PR is welcome

JustFly1984:
@polys  please check latest version 2.10.1

PR #2994, opened 2022-04-17
Prevent 'Google Sans Text' font download

polys:
It appears that Google Maps started making requests for the `Google Sans Text` font, in addition to/instead of `Roboto`.

Just trying to make `preventGoogleFontsLoading` behave as expected — see #2981


JustFly1984 (approved):
Thank you!

PR #3007, opened 2022-04-26
React hooks are available in React 16.8 through 18

Downchuck:
Currently the main branch specifies React 18+ as a peer dependency. The issue comes up when running `npm install` as one must use `--legacy-peer-deps` to override the present limited peerDependency if using an earlier version of React. This patch extends the package.json specification to include 16.8 and 17, which both support hooks and both run the project.


JustFly1984 (approved):

JustFly1984:
released 2.10.1

PR #3016, opened 2022-05-04
patch(optional_options): Allowing MarkerClusterer options to be optional

addy:
# Please explain PR reason
Google's js-markerclusterer lib allows for completely optional arguments during construction. This will default to the SuperClusterAlgorithm.

https://github.com/googlemaps/js-markerclusterer/blob/69d026ebfca8e3732e3088e234a450409fd63a11/src/markerclusterer.ts#L77


JustFly1984:
this will break the code, due to imposibility to spread undefined down the code flow.

Issue #3022, opened 2022-05-11
SIngle click event on MarkerClusterer stopped working


aritrasyd:
## Your Environment

os: mac windows

node --version 16.14.2

react version 17.0.2

@react-google-maps/api version 2.1.1

The single click on the cluster does not work any more with the lastest package build using the version 2.1.1 or higher. The single click still works if I use the previously installed node modules. Using the below codesandbox.com url, the same issue can be reproduced using version 2.8.1 onwards. Now, user needs to double click to zoom on the cluster.

## Basic implementation of incorrect behavior in codesandbox.com

https://codesandbox.io/s/marker-clusterer-api-forked-ex4tmj?file=/src/Map2.js:552-562

kevindoole:
Ditto! I perused the source but couldn't find a solution. I noticed a couple possible angles that may be clues, but didn't confirm:

- the `MarkerClusterer` `options` prop seems to have some overlap with it's props in the docs. Eg. `imagePath` is given is an option, but it's also in the list of props
- in @react-google-maps/marker-clusterer, they have an option called `onClusterClick` -- perhaps `onClick` is mixed up with `onClusterClick`
- the `zoomOnClick` functionality is also not working

pedromedley:
Hey there! Just submitted a PR to fix this issue. The problem is that the event listeners in the ClusterIcon should be added to the div and not via google.maps.event.addListener according to the deprecation message of the google.maps.event. addDomListener.

JustFly1984:
I've released 2.10.3. fixed memory leaks

pedromedley:
@JustFly1984 now there is a problem where in the new event handler functions, the this is referencing to the div and not to ClusterIcon. I will submit a new PR to fix this and include 2 additional changes:
1. the ClusterIconStyle.className is not applied to the cluster div on the first draw, only after the ClusterIcon is re-added -> will assign the className to the div in the  show() function instead on the add()
2. adding the width and the height properties to the img so it will match the ClusterIcon.width and ClusterIcon.height -> this will allow to provide images with higher resolution so they can look crispier

pedromedley:
@JustFly1984 just created a new PR

JustFly1984:
Can we close an issue now? Please test latest release 2.11.8

kevindoole:
Confirmed working for me too -- thanks!!! ❤️ 🏆

aritrasyd:
Thanks for the fix. The latest version worked for me ❤️

[...]

PR #3025, opened 2022-05-18
Fixing a problem where the click event on the cluster is not fired

pedromedley:
This change will add the event listeners to the marker div element instead of using the google.maps.event.addListener().

This change should fix [#3022](https://github.com/JustFly1984/react-google-maps-api/issues/3022)

# Please explain PR reason
The deprecation message states what should be used instead:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6963d81399600abb1a2ee621705a44b8b67cf212/types/google.maps/index.d.ts#L6963


JustFly1984 (approved):

JustFly1984:
released 2.10.3

PR #3052, opened 2022-07-03
fix interface GroundOverlayProps - bounds

aeromosaico:
# Please explain PR reason
bounds interface was not working


JustFly1984:
I guess what you have done is incomplete fix. yu can't just rename google.maps.LatLngBounds into google.maps.LatLngBoundsLiteral

Your current change can break types for people who are passing `google.maps.LatLngBounds` as bounds already.

you need to change it to `google.maps.LatLngBoundsLiteral | google.maps.LatLngBounds` so it satisfies the both options availabe to pass to `new google.maps.GroundOverlay()` as second argument.
 Please make changes, and I will merge and release.

PR #3081, opened 2022-08-22
Fix onCloseClick of InfoWindowF not triggering

haihala:
# Please explain PR reason
Using @react-google-maps/api for a project and for some reason, InfoWindowF onCloseClick is not firing. Went through the code and this seems like an obvious fix, althought I'm not sure if it's the only change required.


JustFly1984 (approved):

JustFly1984:
@haihala released 2.21.2

PR #3106, opened 2022-09-22
Added OverlayViewF as a Functional Component

### LINKED ISSUE #3080: On drag end the OverLayView tag  flickering
# Issue template

The new react version 18 having some issues with the react google maps api on drag end.

https://codesandbox.io/s/googlemapreact-demo-forked-5gdsib

## Please provide an explanation of the issue

We were working in this issue the last days and we found that it is an issue between React 18 and the maps library because as soon you downgrade the react version to the 17 the flickering issue disappears 

## Your Environment

os: mac/

node --version 14.7.1

react version 18

webpack version

@babel version

@react-google-maps/api version 2.12.1

## How does it behave?

When you drag en finish this the overlayview content flickering.

The overlayview content should stay in the same spots where the developer place it 

## Basic implementation of incorrect behavior in codesandbox.com

This is not my code but is some one code who has the same issue 

https://codesandbox.io/s/googlemapreact-demo-forked-5gdsib

andrewdoro:
# Please explain PR reason
Added OverlayViewF as  Functional Component. Adapted code from Daw-Chih Liou blog post.
Fixes #3080

### DIFF
--- a/packages/react-google-maps-api/src/components/dom/OverlayView.tsx
+++ b/packages/react-google-maps-api/src/components/dom/OverlayView.tsx
@@ -4,37 +4,62 @@ import invariant from 'invariant'
 
 import MapContext from '../../map-context'
 
-import { getOffsetOverride, getLayoutStyles, arePositionsEqual } from './dom-helper'
-import { ReactNode, CSSProperties, PureComponent, RefObject, createRef, ReactPortal, Children } from 'react'
+import {
+  getOffsetOverride,
+  getLayoutStyles,
+  arePositionsEqual,
+} from './dom-helper'
+import {
+  ReactNode,
+  CSSProperties,
+  PureComponent,
+  RefObject,
+  createRef,
+  ReactPortal,
+  Children,
+  useContext,
+  useMemo,
+  useEffect,
+} from 'react'
+import { createOverlay } from './Overlay'
 
 interface OverlayViewState {
   paneEl: Element | null
   containerStyle: CSSProperties
 }
 
-function convertToLatLngString(latLngLike?: google.maps.LatLng | google.maps.LatLngLiteral | null) {
+function convertToLatLngString(
+  latLngLike?: google.maps.LatLng | google.maps.LatLngLiteral | null
+) {
   if (!latLngLike) {
     return ''
   }
 
-  const latLng = latLngLike instanceof google.maps.LatLng
-    ? latLngLike
-    : new google.maps.LatLng(latLngLike.lat, latLngLike.lng)
+  const latLng =
+    latLngLike instanceof google.maps.LatLng
+      ? latLngLike
+      : new google.maps.LatLng(latLngLike.lat, latLngLike.lng)
 
   return latLng + ''
 }
 
-function convertToLatLngBoundsString(latLngBoundsLike?: google.maps.LatLngBounds | google.maps.LatLngBoundsLiteral | null) {
+function convertToLatLngBoundsString(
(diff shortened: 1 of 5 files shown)


packages/react-google-maps-api/src/components/dom/OverlayView.tsx
```diff
@@ -44,14 +69,69 @@ export interface OverlayViewProps {
...
+  zIndex?: number
   onLoad?: ((overlayView: google.maps.OverlayView) => void) | undefined
   onUnmount?: ((overlayView: google.maps.OverlayView) => void) | undefined
 }
+export const FLOAT_PANE: PaneNames = `floatPane`
+export const MAP_PANE: PaneNames = `mapPane`
+export const MARKER_LAYER: PaneNames = `markerLayer`
+export const OVERLAY_LAYER: PaneNames = `overlayLayer`
+export const OVERLAY_MOUSE_TARGET: PaneNames = `overlayMouseTarget`
+export function OverlayViewF({
```
JustFly1984, line 89: please use memo for export like other F components

JustFly1984 (changes requested):

packages/react-google-maps-api/src/components/dom/OverlayView.tsx
andrewdoro: added memo

JustFly1984 (approved):

packages/react-google-maps-api/src/components/dom/OverlayView.tsx
```diff
@@ -44,14 +70,71 @@ export interface OverlayViewProps {
   children?: ReactNode | undefined
   // required
   mapPaneName: PaneNames
-  getPixelPositionOffset?: ((offsetWidth: number, offsetHeight: number) => { x: number; y: number }) | undefined
-  bounds?: google.maps.LatLngBounds | google.maps.LatLngBoundsLiteral | undefined
-  position?: google.maps.LatLng | google.maps.LatLngLiteral | undefined
+  position: google.maps.LatLng | google.maps.LatLngLiteral
```
joshuapaling, line 73: Was there a reason why position went from optional to required here?

dogofpavlov: No clue either... I basically asked the same question and instead of an answer was told to make a PR for 1 character change. I just downgraded.

joshuapaling: It's not just a 1 line change to fix. The problem isn't just that the ` | undefined` was taken off, it's that the existing `OverlayView` component allowed you to pass either a `position` or a `bounds`. The new `OverlayViewF` still accepts a `bounds` as optional - but doesn't do anything with it. And that's why `position` needs to be mandatory.

So, a use case - specifying a bounding box - is no longer supported. This PR probably shouldn't have been merged.

joshuapaling: @andrewdoro was the decision to not support `bounds` an oversight or intentional?

andrewdoro: I think it was an oversight, I didn't think that bounds could be used like that. The bounds logic should be added it here to make it work
https://github.com/andrewdoro/react-google-maps-api/blob/b7a9401a2ecddbf27df9ec961ee3340688a408a8/packages/react-google-maps-api/src/components/dom/Overlay.tsx#L13-L36

PR #3122, opened 2022-11-02
Fix onDragEnd useEffect dependency

applepenguin:
# PR reason
map listenner is not updated because onDblClick is specified in the dependency of onDragEnd.

## changes
Fixed onDragEnd's useEffect dependency to specify onDragEnd


ivannovazzi (approved):
LGTM 👍

JustFly1984 (approved):

Issue #3114, opened 2022-10-11
the latest version 2.13 included a breaking change for OverlayView


dogofpavlov:
Using the latest 2.13 version it included a change that made the "position" property of "OverlayViewProps" required. Previously this property was optional.  I am using the OverlayView by passing only a "bounds" property of LatLngBounds. Is it a new requirement to also pass the position? I could get the center of the bounds and use that but I'm not sure what other effects this would have.

fireboltfrog:
If this is not a bug, the position property should also be labeled as required in the documentation.

fireboltfrog:
Definitely looks like a bug to me. 

@dogofpavlov did you manage the OverlayView to consider the bounds property as expected?

JustFly1984:
@fireboltfrog @dogofpavlov  you PR is welcome

joshuapaling:
Looks like this is the line that did it: https://github.com/JustFly1984/react-google-maps-api/pull/3106/files#diff-12da235eb26b2571f4bc4a5a0b1686d9c1462b454f6b297afb910572a41bd5f3R73

PR #3154, opened 2022-12-01
Fix backward incompatible change in OverlayViewF

joshuapaling:
As per https://github.com/JustFly1984/react-google-maps-api/pull/3106#pullrequestreview-1198656139, a breaking change was introduced where `OverlayViewF` stopped supporting the `bounds` prop, and made `position` required.

It also changed the way styles are set from assigning width/height, to using transform - though I'm not aware of any bugs that created.

This PR makes the `OverlayViewF` support the same features as `OverlayView` and closes https://github.com/JustFly1984/react-google-maps-api/issues/3114

See GIF of storybook - the red boxes already existed, the yellow one is new, using the `bounds` prop.

WithBounds

Cutting a patch release would be much appreciated! If that won't happen soon please LMK and I'll use my fork in the meantime.

cc @andrewdoro you may also want to review this.

### DIFF
--- a/packages/react-google-maps-api/src/components/dom/Overlay.tsx
+++ b/packages/react-google-maps-api/src/components/dom/Overlay.tsx
@@ -1,4 +1,4 @@
-import { getOffsetOverride } from './dom-helper'
+import { getOffsetOverride, getLayoutStyles } from './dom-helper'
 
 type fnPixelPositionOffset = (
   offsetWidth: number,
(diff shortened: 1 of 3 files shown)


packages/react-google-maps-api/src/components/dom/Overlay.tsx
```diff
@@ -7,23 +7,27 @@ type fnPixelPositionOffset = (
 export function createOverlay(
   container: HTMLElement,
   pane: keyof google.maps.MapPanes,
-  position: google.maps.LatLng | google.maps.LatLngLiteral,
+  position?: google.maps.LatLng | google.maps.LatLngLiteral,
```
JustFly1984, line 10: Forgot to add undefined to optional prop

JustFly1984 (approved):

joshuapaling:
Thanks for merging! Any ETA on when you'll cut a release?

PR #3186, opened 2023-01-27
refactor: Ensure `react/jsx-runtime` isn't inlined into bundle

rschristian:
Currently, `react/jsx-runtime` is being inline into your bundles due to it not being marked as external. You can plainly see this in the built bundles, [React's license acts as an easy to spot annotation](https://www.runpkg.com/?@react-google-maps/api@2.17.1/dist/esm.js#14).

This bloats the bundle and makes it incompatible with Preact (see https://github.com/preactjs/preact/issues/3870), as React cannot be aliased out if it's being inlined.


JustFly1984 (approved):

rschristian:
Thanks!

PR #3191, opened 2023-01-30
Updating its own load state if it was already loaded externally

Alynva:
If the `LoadScript` sees the Google API already loaded, it should proceed to mount its children components. Right now, it gets stuck in the loading state, showing just the console message. Instead, it needs to consider its work 'done'.


JustFly1984:
Sorry, but this library does not expect script to be loaded externaly.

PR #3220, opened 2023-03-21
Fix build:docs command

crtl:
- Replaces awesome-typescript-loader with ts-loader
- Excludes **/*.stories.tsx files from styleguidist build
- Fix acorn parse error by adding missing semicolons and slash to self-closing-tag

### DIFF
--- a/packages/react-google-maps-api/package.json
+++ b/packages/react-google-maps-api/package.json
@@ -109,7 +109,8 @@
     "rimraf": "3.0.2",
     "rollup": "2.78.1",
     "rollup-plugin-dts": "4.2.2",
-    "rollup-plugin-terser": "7.0.2"
+    "rollup-plugin-terser": "7.0.2",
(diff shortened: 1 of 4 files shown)


JustFly1984:
@crtl are you continue to work on this, or should I merge it as is?

crtl:
@JustFly1984 I thought you wanted to review it?
You can merge it but I dont think the docs are really correct. Ive just fixed the build:docs command so it does not throw errors anymore because you said you are unable to run it.
Problem is that it now only shows the `[ComponentName]F` components (ie functional implementation) in the docs and not the `[ComponentName]` anymore.
For Example docs will have `MarkerF` component but no `Marker` component.

JustFly1984:
Well, if docs built with react@17+, it can't make sense to use old broken class components...
The other way would be to fix old components to support react@17+... but I simply have no time. Your PR is very welcome.

PR #3260, opened 2023-07-15
release 2.19.0

JustFly1984:
# Please explain PR reason


packages/react-google-maps-api/src/components/dom/OverlayView.tsx
```diff
@@ -206,10 +210,14 @@ export class OverlayView extends PureComponent<
...
     const { left, top, width, height } = this.state.containerStyle
+
     if (!arePositionsEqual(layoutStyles, { left, top, width, height })) {
       this.setState({
         containerStyle: {
-          ...layoutStyles,
+          top: layoutStyles.top || 0,
+          left: layoutStyles.left || 0,
+          width: layoutStyles.width || 0,
+          height: layoutStyles.height || 0,
```
aleksandrlat, line 220: This is a breaking change. Before there were no width and height set. And width and height are 0 now

JustFly1984:
@aleksandrlat  sorry about that, if you can think of better solution, please PR

PR #3272, opened 2023-08-02
fixes build:docs 🤯 

IndianBoyBR:
### Explanation

Looks like someone used [npm-check-updates](https://www.npmjs.com/package/npm-check-updates) or something similar to update all packages. When doing this, deleting `yarn.lock` file is appropriate, and also check if everything is working as it should 😶‍🌫️

I've lost a couple of hours investigating what happened on git history LOL 🥲, hope we can get the live demo website back with this.

### What as changed:

- Rollback `react-docgen` version;
- Delete all `yarn.lock` files;

Alt Text


JustFly1984:
NoEmit is important, cos we are not building library itself with tsc, but with rollup.

Reverting version back is not the option, due to outdated dependencies or react-docgen and conflicts. Need to setup 6.x correctly.

Rollup possible requires major update to.

ezerssss:
https://react-docgen.dev/docs/migrate/v6

Issue #3374, opened 2024-09-06
Polygons are not correctly removed from the map when unmounting


brandensilva:
# Please provide an explanation

Polygons are not properly being torn down when unmounting. There also appears to be no map reference set on the Polygon itself which may be related to the issue. A similar issue was reported for the Marker by @jwalton here: https://github.com/JustFly1984/react-google-maps-api/issues/1848 and subsequently fixed.

Your Environment
os: mac

node --version: v18.20.2
react version: 18.2
@react-google-maps/api version: 2.19.2

# How is it behave?

A visual goes a long ways. This is a zoomed in portion of the map. We select a map marker, loading in the polygons given the context of data, exit that state, and then enter it again. You will see that the polygons are persisted and painted over with the same polygons colors and were never unmounted properly from the previous render.

![2024-09-06 09 36 43](https://github.com/user-attachments/assets/050085a6-6fa1-4c5e-b289-1f29d4f8d89d)

# How should it behave correctly?

One would expect when the Polygon unmounts, it should remove the Polygon from the map.

We plan on mirroring Jason's work and submitting a PR to also apply his fix to the Polygon so it is properly cleaned up and removes the Polygon when it is unmounted.

Issue #1848, opened 2020-12-17
Markers are not always correctly removed from map when Marker unmounts.


jwalton:
## Please provide an explanation

Sometimes when removing a marker from the set rendered by react, the marker gets left behind and is not removed from the map.

## Your Environment

os: mac

node --version: v12.19.1
react version: 16.13.1
webpack version: 4.44.1
@babel version: 7.11.6
@react-google-maps/api version: 2.0.2

## How is it behave?

[Here's a quick demo](https://codesandbox.io/s/stupefied-worker-8vqt8?file=/src/Map.tsx) which renders a bunch of points using `supercluster`.  This is (very nearly) following the [supercluster tutorial for google maps](https://www.leighhalliday.com/google-maps-clustering).

supercluster gives us a set of points to render.  At the start, we're zoomed out enough that we just render one:

<img width="443" alt="image" src="https://user-images.githubusercontent.com/1771003/102547891-db5d0080-4087-11eb-9691-e4f0b399cae0.png">

Note the "Rendering 1 markers" at the top of the page.  As we zoom in, supercluster adds more markers.  Here we're zoomed in 4 steps:

<img width="383" alt="image" src="https://user-images.githubusercontent.com/1771003/102548025-0c3d3580-4088-11eb-8622-8da7758a781b.png">

Then when we zoom back out, supercluster removes markers again, until we're back at 1:

<img width="396" alt="image" src="https://user-images.githubusercontent.com/1771003/102548071-195a2480-4088-11eb-8073-8b1202173344.png">

But, there is obviously more than just one marker being rendered.  The problem is that react-google-maps-api is leaking markers - markers are unmounting but are not being removed from the map.

You can also verify this with a quick change to [Marker.tsx](https://github.com/JustFly1984/react-google-maps-api/blob/develop/packages/react-google-maps-api/src/components/drawing/Marker.tsx).  In `componentDidMount()`:

```diff
    const marker = new google.maps.Marker(markerOptions)
+    this._marker = marker
```

And in `componentWillUnmount()`:

```diff
  componentWillUnmount(): void {
    if (this.state.marker !== null) {
      if (this.props.onUnmount) {
        this.props.onUnmount(this.state.marker)
      }

      unregisterEvents(this.registeredEvents)

      if (this.props.clusterer) {
        this.props.clusterer.removeMarker(
          // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
          // @ts-ignore
          this.state.marker,
[...]

PR #3375, opened 2024-09-06
fix: alter state to allow polygon to tear down on unmount, setMap reference on polygon

brandensilva:
This fixes this issue: https://github.com/JustFly1984/react-google-maps-api/issues/3374 and applies a similar solution that @jwalton had come up with for the Marker issue: https://github.com/JustFly1984/react-google-maps-api/issues/1848.

This allows the Polygon to properly unmount and sets a map reference on the polygon.


JustFly1984 (approved):

JustFly1984:
You can use PolygonF component

JustFly1984:
released 2.20.1

PR #3380, opened 2024-09-24
Update Polygon onEdit to include remove_at listener

JacobGalati:
Currently when editing a Polygon if you add a new point and and then click the undo button the `onEdit` event does not fire indicating the return to the original path because it is a `remove_at` event. However if you move an existing point and then click the undo button the `onEdit` event _does_ fire since it is a `set_at` event. This is an inconsistent behaviour.

This is fixed by adding a `remove_at` listener to the Polygon's path in the same way as the `insert_at` and `set_at` listeners.


JustFly1984 (approved):

JustFly1984:
released 2.20.1

PR #3382, opened 2024-10-06
Prepare for React 19

memark:
This PR updates the code to eliminate some type problems with React 19 RC.

- Use `React.JSX` instead of global `JSX.`
- Use `React.ReactNode` instead of `React.ReactChild`

PR #3388, opened 2024-10-12
Fix lint error

memark:
Fixes this lint error

image

just introduced in #3382.


JustFly1984:
current work is happening in upgrade branch.

